PR #17, opened 2017-10-11
Fix subtle bug in resource id extraction

killthrush:
### Description
Fix a subtle nondeterministic bug in resource id extraction - timestamps
were an unintentional factor in the deduping code.  If the code ran
quickly (usually on the order of milliseconds), the timestamps might match
for all extracted resource IDs.  If the code runs more slowly, or straddles
the boundary between epoch ticks, then we might see duplicates sneaking in.

The bug itself would simply cause extra writes and isn't all that serious,
however, it was tripping a few very intentionally sensitive unit tests and
causing errors with CI.

In addition to this bug fix, remove all traces of the Resource Activity
table and Resource Create lambda; these constructs have proven to be too
inefficient to continue using.  Removing them now cleans up the code
in the vicinity of the bug fix significantly.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -60,12 +59,11 @@ class CloudTrailEventConsumer(object):
         _resource_type_regexp,
     ]
```
killthrush: Remove the environments dependency - the env_ids are now part of the normalized events so we don't need this extra code anymore (resolves TODO #1 in here)

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -76,25 +74,8 @@ def __init__(self, environments, ttl=_DEFAULT_TTL):
         self._EXCLUSION_STRINGS = ['eventID', 'sharedEventID', 'eventName', 'eventSource',
                                    'userAgent', 'roleSessionName', 'requestID']
         self._resource_ids = []
-        self.environments = environments
         self.ttl = ttl
```
killthrush: Since we're no longer writing to the resource activity table, we don't need to maintain the attributes and functions used to support writes to this table

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -105,11 +86,7 @@ def extracted_ids(self):
         Returns:
             A de-duped list of resource IDs and paths
```
killthrush: Deduping is now a lot simpler without the resource activity table

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -205,33 +182,20 @@ def _extract_primary_resource_id(self, normalized_event: dict) -> str:
...
-        # TODO: a number of these items support the resource activity table, which is being phased out.
-
         service_type = normalized_event['event_source'].split('.')[0]
         account_id = normalized_event['aws_account_id']
         event_type = normalized_event['event_name']
-        cz_res_id = self._create_cz_res_id(account_id, aws_res_id_value, service_type)  # deprecated field
         resource_tracking_id = self._create_tracking_id(account_id, event_type, aws_res_id_path, aws_res_id_value)
-        environment = self.environments.get(account_id)  # TODO: this will be in normalized events soon
-        current_timestamp = int(parse(datetime.now(timezone.utc).isoformat()).timestamp())
         metadata = {
```
killthrush: Remove all the deprecated fields.  What's left are the fields written to the tracking table

reactor/discovery/create_resource_handler.py
```diff
@@ -1,87 +0,0 @@
-# -*- coding: utf-8 -*-
```
killthrush: We stopped using this lambda a few weeks ago due to performance concerns.  It does not appear like we will ever return to it, and it depended on the Activity table so I just got rid of it now

reactor/discovery/receive_events_handler.py
```diff
@@ -65,7 +65,6 @@ def receive_events(event, context):
         current_account_id = aws_tools.get_caller_account_id()
         environments = dynamodb.CZReactorEnvironments(ambient_storage.namespace)
```
killthrush: Remove code that uses the activity table

reactor/serverless.yml
```diff
@@ -95,15 +95,6 @@ custom:
         minimum:
           prod: 5
           default: 1
-    CzReactorResourceActivity:
```
killthrush: Remove activity table and Create lambda

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -76,25 +74,8 @@ def __init__(self, environments, ttl=_DEFAULT_TTL):
...
-        Returns:
-            A de-duped list of resource IDs and timestamps
-        """
-        keys_to_filter = {'resource_tracking_id', 'created_timestamp', 'last_updated_timestamp',
-                          'expiration_timestamp', 'source_path', 'source_input', 'source_operation',
-                          'aws_res_id', 'times_seen', 'disposition'}
-        filtered_ids = [{k: v for k, v in f.items() if k not in keys_to_filter} for f in self._resource_ids]
-        deduped_list = [dict(t) for t in set([tuple(d.items()) for d in filtered_ids])]
-        return deduped_list
+        self.current_timestamp = int(parse(datetime.now(timezone.utc).isoformat()).timestamp())
```
imcgaunn: why do we have to write the date as an ISO formatted string and then parse it again?

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -205,33 +182,20 @@ def _extract_primary_resource_id(self, normalized_event: dict) -> str:
...
+            'cz_env_id': normalized_event['environment_id'],
+            'created_timestamp': self.current_timestamp,
+            'last_updated_timestamp': self.current_timestamp,
+            'expiration_timestamp': self.current_timestamp + self.ttl,
             'source_path': aws_res_id_path,
-            'aws_res_id_path': aws_res_id_path,  # deprecated field
-            'aws_res_id_key': aws_res_id_path.split('.')[-1],
-            'aws_res_id_value': aws_res_id_value,  # deprecated field
-            'aws_res_id': aws_res_id_value,  # deprecated field
+            'aws_res_id': aws_res_id_value,
```
imcgaunn: this appears to have been previously marked as a deprecated field. Do we need this now?

killthrush: good eye - `aws_res_id` was marked as deprecated in error.  I found it while fixing the bug

reactor/discovery/cloudtrail_event_consumer.py
killthrush: that is an excellent question =)

I will see if I can simplify this

test/unit/discovery/test_cloudtrail_event_consumer.py
```diff
@@ -29,27 +29,11 @@ def mock_event_mapping():
 
 @pytest.fixture
 def mock_environments():
```
killthrush: This bit prevents mock references from being used in the event output

reactor/discovery/cloudtrail_event_consumer.py
killthrush: simplified

PR #19, opened 2017-10-12
Feature/reactor 28

imcgaunn:
### Description

Add a `configure-unregister-account` function that acts as an inverse of `configure-register-account`, removing the SNS subscription from the ingest function and deleting the environment settings associated with account.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/database/models/dynamodb/cz_reactor_environments.py
```diff
@@ -74,6 +74,23 @@ def get(self, account_id):
...
+            success (bool): success of delete operation
+        """
+        self._clear_caches()
+        try:
+            self.table.delete_item(Key={'account_id': account_id})
+            return True
+        except botoerr.ClientError as err:
+            logger.debug('Unable to delete env associated with account_id {}: {}'.format(account_id, err))
+            return False
+
```
killthrush: Would log a warning instead on line 91

reactor/database/models/dynamodb/cz_reactor_environments.py
```diff
@@ -74,6 +74,23 @@ def get(self, account_id):
...
+        """ remove the environment associated with account_id. This operation
+        is destructive, and if successful, information for the environment
+        will be removed permanently.
+        Args:
+            account_id (str): aws account id
+        Ret:
+            success (bool): success of delete operation
+        """
+        self._clear_caches()
+        try:
```
killthrush: If `account_id` is an int, will this still work?  If not, should we coerce the type?

reactor/handlers/configure.py
```diff
@@ -236,6 +236,58 @@ def subscribe_reactor_function(account_details, ingest_function, namespace):
...
+        lambda_client: boto3 lambda client.
+
+    Returns:
+        (bool): True if unsubscription is successful, else False.
+    """
+
+    function_name = function_info['FunctionName']
+    function_arn = function_info['FunctionArn']
+
+    topic_subs = sns_client.list_subscriptions_by_topic(TopicArn=sns_arn)['Subscriptions']
```
killthrush: on 261, recommend adding test to say that this is a list of subscriptions

reactor/handlers/configure.py
```diff
@@ -236,6 +236,58 @@ def subscribe_reactor_function(account_details, ingest_function, namespace):
     return True
 
 
+def unsubscribe_reactor_function(sns_arn, function_info, namespace, sns_client, lambda_client):
+    """
+    Given an SNS ARN, function information in the format returned by the boto3 library's
+    Lambda.Client.list_functions method, a reactor namespace and boto3 clients for accessing
+    the necessary services (SNS and lambda), unsubscribes SNS from the reactor's ingest function
+    and removes the policy allowing SNS to invoke it.
+
```
killthrush: OK, so 'function' in this case mainly refers to ingest right?  Given the contents of the code, I might suggest making this more explicit in the args and the docstring.  I don't think it's worth trying to make this generic

reactor/handlers/configure.py
```diff
@@ -236,6 +236,58 @@ def subscribe_reactor_function(account_details, ingest_function, namespace):
...
+    """
+
+    function_name = function_info['FunctionName']
+    function_arn = function_info['FunctionArn']
+
+    topic_subs = sns_client.list_subscriptions_by_topic(TopicArn=sns_arn)['Subscriptions']
+    logger.info('{}'.format(topic_subs))
+    ingest_subs = [s for s in topic_subs
+                   if s['Endpoint'] == function_arn]
+
```
killthrush: Should 267 be a warning?  I can't decide between that and `info`.  Is this a case that really shouldn't happen?

reactor/handlers/configure.py
```diff
@@ -236,6 +236,58 @@ def subscribe_reactor_function(account_details, ingest_function, namespace):
...
+    # If no subscriptions returned, we've already unsubscribed (or it was never subscribed).
+    if not ingest_subs:
+        logger.info('Ingest function is not currently subscribed to sns topic: {}'.format(sns_arn))
+
+    if len(ingest_subs) > 1:
+        logger.warning('ingest function {} subscribed to topic {} multiple times')
+
+    for s in ingest_subs:
+        try:
```
killthrush: 276 probably should be `info` instead of `debug`

reactor/handlers/configure.py
```diff
@@ -236,6 +236,58 @@ def subscribe_reactor_function(account_details, ingest_function, namespace):
...
+            logger.debug('unsubscribed from arn: {}'.format(arn))
+        except botoerr.ClientError as e:
+            logger.warning('unable to subscribe from subscription: {}'.format(arn))
+            return False
+
+        # remove policy allowing SNS to invoke ingest function
+        try:
+            lambda_client.remove_permission(FunctionName=function_name,
+                                            StatementId='cloudzero-reactor-{}'.format(namespace))
+        except botoerr.ClientError as e:
```
killthrush: Maybe use `warning` on 286?  or, if you keep `error`, 278 probably should be `error` as well

reactor/handlers/configure.py
imcgaunn: This makes sense

reactor/handlers/configure.py
imcgaunn: Good catch, meant to add such text and never got back to it.

reactor/handlers/configure.py
```diff
@@ -335,6 +387,69 @@ def _register(event, context, **kwargs):
...
+        return resp.generate_response(400, json.dumps({'error': 'Invalid event: body not valid json.'}))
+
+    account_details = event_map.get('account_details')
+    if not account_details:
+        return resp.generate_response(400, json.dumps({'error': 'Invalid event: body does not have account details.'}))
+
+    sns_arn = account_details.get('sns_arn')
+    if not sns_arn:
+        return resp.generate_response(400, json.dumps({'error': 'Invalid account details: missing SNS ARN.'}))
+
```
killthrush: Is there a catch-all somewhere to deal with dynamodb exceptions, etc?

reactor/handlers/configure.py
imcgaunn: I'll use a `warning` on 286.

reactor/handlers/configure.py
```diff
@@ -335,6 +387,69 @@ def _register(event, context, **kwargs):
...
+    environments = dynamodb.CZReactorEnvironments(namespace)
+    acct_id = account_details.get('connected_account_id')
+    env_settings = environments.get(acct_id)
+    reactor_functions = get_reactor_functions(namespace)
+    ingest_function = reactor_functions['cloudzero-reactor-{}-ingest-event'.format(namespace)]
+
```
killthrush: Hmm, I'm wondering if we should move the function name template into the ingest handler module... Do we need this elsewhere?

reactor/handlers/configure.py
```diff
@@ -335,6 +387,69 @@ def _register(event, context, **kwargs):
...
+    # NB: A service has to be set here due to Provider's laziness.
+    # Provider will not create a boto3 session until a service type has
+    # been set. This feels unnatural here, where all we need the provider
+    # for is to create a session, so we can create more clients of other
+    # service types.
+    aws_provider = provider.Provider(service='sns')
+    sns_client = aws_provider.session.client('sns')
+    lambda_client = aws_provider.session.client('lambda')
+
```
killthrush: re: your comment on Provider - would it be cleaner just to use multiple Provider instances instead of just one?

reactor/handlers/configure.py
imcgaunn: I think it should be a warning. It normally shouldn't happen, but in the case that it does, I would like `disconnect` to continue removing the resources it can.

reactor/handlers/configure.py
```diff
@@ -335,6 +387,69 @@ def _register(event, context, **kwargs):
...
+    unsubscribed = unsubscribe_reactor_function(sns_arn, ingest_function, namespace, sns_client, lambda_client)
+    logger.debug('sns unsubscribed status: {}'.format(unsubscribed))
+    settings_removed = True if not env_settings else environments.remove(acct_id)
+    logger.debug('account removed status: {}'.format(settings_removed))
+
+    checks = [{'name': 'sns subscription', 'success': unsubscribed},
+              {'name': 'environment settings', 'success': settings_removed}]
+    failed_checks = [c for c in checks if not c['success']]
+    success = all(checks)
+
```
killthrush: Kind of puzzled by the `201 - Created` here.  Should this be 200 instead?

reactor/database/models/dynamodb/cz_reactor_environments.py
imcgaunn: in the table's schema the 'account_id` key is defined as a string. I can coerce `int`'s to string, should the account_id end up being an `int`.

reactor/handlers/configure.py
imcgaunn: It would be cleaner, but I wanted to avoid creating multiple boto sessions unnecessarily. That being said, this may be a premature optimization, so I will use multiple `Provider` instances.

reactor/handlers/configure.py
imcgaunn: That's a good idea; we do need this elsewhere.

reactor/handlers/configure.py
imcgaunn: I wasn't sure what to do here. I thought that since resources were changed as a result of this API that it might make sense to use a `201`, but if that doesn't jive with the semantics of `201` status codes I will change it `200`.

killthrush: Yeah I would prefer to use 200 here.  201 is pretty specialized

PR #20, opened 2017-10-15
Feature/reactor 156

killthrush:
### Description
Implement v1 of the discovery curate function.  It is hooked up to
dynamodb streams and will execute when the Resource Tracking table
changes.

Specifically, it will:
* Send INSERT records to various plugins based on service type. If the
  service type is not supported (no plugin exists), then the resource
  record is set to 'not_supported' until it expires.
* Add REMOVE records back to the tracking table, i.e. "refreshing"
  older records whose TTL has expired (or have been removed manually).
* Ignore MODIFY records.  Curation is driven off inserted records only.

Also add covering unit tests.

Also in this PR:
* Move reactor event filtering to CloudtrailEventConsumer
* Add a new lambda decorator to reduce boilerplate in code and tests
* Enable localstack support for lambda
* Temporarily disable counting of 'times seen" for resource IDs (takes pressure of dynamodb streams)
* Fix a number of small bugs and style issues 

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/constants.py
```diff
@@ -14,9 +14,6 @@
 NOT_OK = "NO"
 MAYBE_OK = "MAYBE"
```
killthrush: These are no longer used per a change made long ago

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -12,18 +12,10 @@
 import hashlib
 from datetime import datetime, timezone
 from utils import search_keys, setup_logging, advanced_dict_search, snake_case_split, camel_case_split
+from discovery.constants import DISPOSITION_NEW, SOURCE_CLOUDTRAIL
 
 logger = setup_logging('reactor')
```
killthrush: Move these to a discovery-centric constants file.  they are used in a number of places ATM (might be a code smell?)

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -43,6 +35,11 @@ class CloudTrailEventConsumer(object):
     _principal_regexp = re.compile(r'^[A-Z0-9]{21}$')
     _resource_type_regexp = re.compile(r'^[a-zA-Z0-9]+::[a-zA-Z0-9]+::[a-zA-Z0-9]+$')
```
killthrush: The filtering code didn't change, but it was moved here from being inside the Resource Receive handler

reactor/discovery/curate_resource_handler.py
```diff
@@ -0,0 +1,113 @@
+# -*- coding: utf-8 -*-
```
killthrush: This is brand new.  Hopefully this will be easy to follow given the comments and the unit tests

reactor/discovery/facade/resource_activity_table.py
```diff
@@ -1,67 +0,0 @@
-# -*- coding: utf-8 -*-
```
killthrush: We stopped using this code in a previous ticket, but forgot to clean up

reactor/discovery/facade/resource_tracking_table.py
```diff
@@ -46,10 +49,10 @@ def write(self, records):
                 logger.error('Failed to write record due to type mismatch')
                 logger.error(pprint.pformat(data))
                 raise ValueError('Failed to write record due to type mismatch: {}'.format(data))
```
killthrush: Found a cleaner way to deal with boto3 exceptions.  The downside is that it's not always consistent; ValidationException is missing from the botocore metadata for some reason

reactor/discovery/receive_events_handler.py
```diff
@@ -3,35 +3,26 @@
...
 
-ambient_storage = AmbientLambdaStorage()
+ambient_storage = lambda_utils.AmbientLambdaStorage()
 logger = utils.setup_logging('reactor')
 
-# This is the key of the actor account in a normalized cloudtrail event message.
-# The actor is the party that triggers an event; this information is typically found
-# via the event-mapping process because it can vary from service to service.
-__ACTOR_ACCOUNT_EVENT_KEY = 'event_actor_account_id'
-
```
killthrush: most of the changes here are either in removing the boilerplate and applying the decorator, or it's moving the reactor account filtering code

reactor/discovery/receive_events_handler.py
```diff
@@ -51,77 +42,52 @@ def receive_events(event, context):
...
+                                                                    current_account_id))
+
+    # Only attempt a record insert if we haven't already done one.  This is mainly for cost savings.
+    extracted_ids = {item['resource_tracking_id'] for item in consumer.extracted_ids}
+    parameters = [{'resource_tracking_id': id} for id in extracted_ids]
+    known_items = resource_tracking_dbh.read_multiple(parameters)
+    known_ids = {item['resource_tracking_id'] for item in known_items}
+    items_to_write = [item for item in consumer.extracted_ids if item['resource_tracking_id'] not in known_ids]
+    resource_tracking_dbh.write(items_to_write)
+
```
killthrush: Important note here.  Eventually we will need this feature again, but I'm not happy with the current implementation.

reactor/serverless.yml
```diff
@@ -252,6 +252,15 @@ functions:
     handler: discovery/receive_events_handler.receive_events
     events:
       - s3: ingest
```
killthrush: setting up streams in serverless is easy peasy

reactor/utils/lambda_utils.py
```diff
@@ -81,6 +81,27 @@ def configured(self):
         return bool(self.namespace) and bool(self.settings)
```
killthrush: This function is super handy!  I wasn't able to find any guidance on how to do this, but looking through the boto3 source proved useful

test/unit/discovery/plugins/s3/test_enrich_resource_handler.py
```diff
@@ -92,28 +89,11 @@ def test_aws_general_client_error(lambda_module, valid_lambda_environment, mock_
     call_handler = lambda_module.enrich_resource
     call_handler(generic_enhance_event, {})
     verify_log('ClientError', mock_logger.exception, expected_call_count=1, which_call=0)
-    verify_log('super_bucket', mock_logger.error, expected_call_count=3, which_call=1)
-
```
killthrush: Moving boilerplate to the decorator makes handler tests a lot simpler.  These tests still exist, but they are testing the decorator features.

test/unit/discovery/test_curate_resource_handler.py
```diff
@@ -0,0 +1,344 @@
...
+import attrdict
+import logging
+import test.mock_helpers as test_helpers
+import test.unit.mock_helpers as unittest_helpers
+import discovery.curate_resource_handler as handler
+import utils.lambda_utils as lambda_utils
+import discovery.facade as facade
+import datetime
+
+
```
killthrush: Note this style of fixture for setting up mocks - it's pretty close to ideal in terms of simplicity (though it might be possible to reduce code with some helper functions).  There is only one dependent fixture (`mocker`, part of pytest-mock), and all mocks are combined into a single struct.  You don't need to go looking for this arrange/setup code except here.

test/unit/discovery/test_curate_resource_handler.py
```diff
@@ -0,0 +1,344 @@
...
+    symbol = '{}.datetime'.format(prefix)
+    context.clock = mocker.patch(symbol, autospec=datetime)
+    context.clock.datetime.now.return_value = datetime.datetime(2007, 1, 2, tzinfo=datetime.timezone.utc)
+
+    yield context
+    mocker.stopall()
+    os.environ = orig_env
+
+
```
killthrush: As a result of using the `context` fixture, the tests are really clean.

test/unit/discovery/test_curate_resource_handler.py
```diff
@@ -0,0 +1,344 @@
...
+    expected_records = [{
+        'resource_tracking_id': 'arbitrary-tracking-id',
+        'service_type': 's3',
+        'disposition': 'refreshed',
+        'created_timestamp': expected_ts,
+        'last_updated_timestamp': expected_ts,
+        'expiration_timestamp': expected_ts + 100,
+        'monkey': 'banana',
+    }]
+
```
killthrush: I like these style of helper functions to group together assert code so it's easier to follow and takes up fewer lines

test/unit/discovery/test_curate_resource_handler.py
```diff
@@ -0,0 +1,344 @@
...
+import discovery.curate_resource_handler as handler
+import utils.lambda_utils as lambda_utils
+import discovery.facade as facade
+import datetime
+
+
+@pytest.fixture()
+def suite_context(mocker):
+    context = attrdict.AttrMap()
+    prefix = handler.__name__
```
imcgaunn: +1 for this nice use of `__name__`

test/unit/discovery/test_cloudtrail_event_consumer.py
```diff
@@ -91,6 +91,57 @@ def test_resource_id_frequencies(event_data, mock_event_mapping, mock_environmen
 
 
 @pytest.mark.unit
+def test_account_filter_account_match_case(mock_event_mapping, mock_environments):
+    """
```
imcgaunn: great documentation on all of these tests

test/unit/discovery/test_curate_resource_handler.py
```diff
@@ -0,0 +1,344 @@
...
+
+def _create_event(operation, service_type):
+    return json.loads(_EVENT_TEMPLATE.format(**{
+        'operation': operation,
+        'service_type': service_type,
+    }))
+
+
+_EVENT_TEMPLATE = """
+    {{
```
imcgaunn: I feel like I'm seeing double here. Why so many curly braces?

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -140,6 +142,13 @@ def consume_event(self, normalized_event: dict):
         Returns:
             None
         """
+        event_principal_account = normalized_event.get(self._ACTOR_ACCOUNT_EVENT_KEY)
+        has_account = self._ACTOR_ACCOUNT_EVENT_KEY in normalized_event
```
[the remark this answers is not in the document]
killthrush: Yep, good catch - I should be able to refactor this away

reactor/discovery/curate_resource_handler.py
```diff
@@ -0,0 +1,113 @@
...
+import datetime
+import utils
+
+ambient_storage = lambda_utils.AmbientLambdaStorage()
+logger = utils.setup_logging('reactor')
+
+_ENRICH_FUNCTION_TEMPLATE = 'discovery-enrich-{}-resource'
+
+
+@decorators.standard_lambda(lambda: ambient_storage, lambda: logger, log_events=False)
```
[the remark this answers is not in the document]
killthrush: The function/thunk pattern is more testable and flexible.  You can choose to form closures around module-level symbols or you could do something else.  Generally speaking, working with module-level variables is inconvenient.  Adam suggested LRU-cached functions to acheive a similar effect.  I think we can iterate on this further, I'm not 100% happy with this solution, but it does make the decorator work the way I want

reactor/discovery/curate_resource_handler.py
```diff
@@ -0,0 +1,113 @@
...
+    Returns:
+        None
+    """
+    logger.debug('Resource Event: {}'.format(event))
+
+    for record in event['Records']:
+        tracking_id = 'Unknown'  # This value is for logging exceptions that can happen before we know tracking ID
+        try:
+            operation = record['eventName']
+            raw_image = record['dynamodb'].get('OldImage', record['dynamodb'].get('NewImage'))
```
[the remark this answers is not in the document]
killthrush: This comes from dynamodb streams: http://docs.aws.amazon.com/lambda/latest/dg/eventsources.html#eventsources-ddb-update.  I think I should probably add that link to the docstring

reactor/discovery/facade/resource_tracking_table.py
```diff
@@ -16,6 +16,9 @@
 
 logger = setup_logging('reactor')
 
+# Keep this many items in the cache for read_item by default
+_DEFAULT_READ_CACHE_SIZE = 512000
```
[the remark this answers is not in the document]
killthrush: Yeah, I want it to cache a LOT of resource IDs.  They are not very big (< 1KB on average), and this *should* fit in memory

reactor/discovery/receive_events_handler.py
```diff
@@ -51,77 +42,52 @@ def receive_events(event, context):
...
+                    'from CloudTrail ingest'.format(len(all_event_records), size_of_payload_in_kb))
+
+    for event_data in all_event_records:
+        consumer.consume_event(event_data)
+    logger.info('Filtered {} events from Reactor account {}'.format(consumer.num_filtered_events,
+                                                                    current_account_id))
+
+    # Only attempt a record insert if we haven't already done one.  This is mainly for cost savings.
+    extracted_ids = {item['resource_tracking_id'] for item in consumer.extracted_ids}
+    parameters = [{'resource_tracking_id': id} for id in extracted_ids]
```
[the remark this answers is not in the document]
killthrush: good idea

reactor/discovery/receive_events_handler.py
```diff
@@ -51,77 +42,52 @@ def receive_events(event, context):
...
+    known_items = resource_tracking_dbh.read_multiple(parameters)
+    known_ids = {item['resource_tracking_id'] for item in known_items}
+    items_to_write = [item for item in consumer.extracted_ids if item['resource_tracking_id'] not in known_ids]
+    resource_tracking_dbh.write(items_to_write)
+
+    # These updates are important, as they track relative frequencies of occurrences of IDs in the cloudtrail
```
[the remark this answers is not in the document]
killthrush: Yes, that was what I was hoping to do eventually.  I'm holding off on the implementation because I'm not quite sure how we want to get this data to core

test/unit/discovery/test_curate_resource_handler.py
killthrush: It's because of the format codes.   hmm, maybe I can do this without format?  I agree this is wonky

reactor/discovery/cloudtrail_event_consumer.py
Tankanow: I thought we were moving away from constants files.

reactor/discovery/cloudtrail_event_consumer.py
```diff
@@ -12,18 +12,10 @@
 import hashlib
 from datetime import datetime, timezone
 from utils import search_keys, setup_logging, advanced_dict_search, snake_case_split, camel_case_split
```
Tankanow: Please import `utils` module and call each function with explicit `utils.` notation.

reactor/discovery/cloudtrail_event_consumer.py
killthrush: Yeah, I feel kinda cheap doing this even though we haven't really made an official stance.  But it's better than importing the lambda handler.  I'll add a comment for this, since it's still bugging me

reactor/discovery/curate_resource_handler.py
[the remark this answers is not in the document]
killthrush: added

reactor/discovery/curate_resource_handler.py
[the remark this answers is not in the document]
Tankanow: Generally, I like @killthrush's approach here. +1 for eliminating module-level variables. The `lambda: ambient_storage` reads nicely. One downside is that the `ambient_storage` and `logger` symbols are only  consistent by convention: line 18 and 19 need to match 24; however the handler body will always use `logger` and `ambient_storage`. I also wonder now how private functions will use the `logger`.

reactor/discovery/curate_resource_handler.py
```diff
@@ -0,0 +1,115 @@
...
+        event (dict): An S3->Lambda event, which contains details about the object that was added.
+                      This is expected to contain one or more S3 keys.
+        context: LambdaContext type containing runtime information from the lambda environment
+
+    Returns:
+        None
+    """
+    logger.debug('Resource Event: {}'.format(event))
+
+    for record in event['Records']:
```
Tankanow: Do we need any input validation on `event` here? Consider `event.get('Records', [])` to short circuit the `for` loop without error.

reactor/discovery/curate_resource_handler.py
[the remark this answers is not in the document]
Tankanow: Why do we prefer the `OldImage` here?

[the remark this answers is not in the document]
Tankanow: Also, consider using `utils.deep_get` here.

reactor/discovery/curate_resource_handler.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _refresh_resource_id(resource_data):
+    """
+    Whenever a resource tracking record expires, we can see it here.  Our reaction is
+    to immediately write it back, which will feed back into the Enrich mechanism.
+    In this way, we can periodically check up on resources from time to time.
+
+    Args:
+        resource_data: (list(dict)) - records to write/refresh
```
Tankanow: The docstring "type" is wrong, below `resource_data` is used just as a `dict` not `list(dict)`.

reactor/discovery/facade/resource_tracking_table.py
```diff
@@ -83,10 +86,13 @@ def read_item(self, tracking_id):
             A dict containing the record data or None if there was no matching record found.
         """
```
Tankanow: I know this was here before, why does this function use `records = []` when it will only ever return 1 item? Simply `return response.get('Items')`.

reactor/discovery/receive_events_handler.py
```diff
@@ -51,77 +42,52 @@ def receive_events(event, context):
...
-            if "ProvisionedThroughputExceededException" in str(err):
-                logger.warning('Exceeded provisioned throughput writing resource data, there is no retry')
-                return
-            else:
-                raise
-    except Exception as err:
-        # We do not log the entire event, because it might just be HUGE
-        logger.error('{}:{}'.format(type(err), err))
-        logger.error(traceback.format_exc())
-        return
```
Tankanow: Consider using `accountid_aware` decorator.

reactor/discovery/receive_events_handler.py
```diff
@@ -51,77 +42,52 @@ def receive_events(event, context):
...
+    bucket_facade = CloudTrailEventBucket(ambient_storage.namespace)
+    current_account_id = aws_tools.get_caller_account_id()
+    consumer = CloudTrailEventConsumer(filtered_account_ids={current_account_id})
+    resource_tracking_dbh = facade.CZReactorResourceTracking(ambient_storage.namespace)
+
+    all_event_records = []
+    start_time = timeit.default_timer()
+    payload_keys = [r['s3']['object']['key'] for r in event['Records']]
```
Tankanow: Consider input validation here on `event`, or maybe just `event.get('Records', [])`.

test/unit/discovery/test_cloudtrail_event_consumer.py
```diff
@@ -91,6 +91,57 @@ def test_resource_id_frequencies(event_data, mock_event_mapping, mock_environmen
...
+    Ensure that events originating from a specified account are filtered.
+    This specifically prevents Discovery from generating events in remote accounts and causing feedback loops.
+    """
+    ct = cloudtrail.CloudTrailEvent(json.loads(ASSUME_ROLE_EVENT4))
```
Tankanow: Were you able to remove the `environments.get` from `ct.normalize` by using the `environment_id` inside the `normalized_event`? If so, you can remove the `mock_environments` references here.

reactor/discovery/cloudtrail_event_consumer.py
killthrush: Will do - no time like the present!

reactor/discovery/curate_resource_handler.py
killthrush: I would kind of like it to fail loudly here.  If the event is malformed, which might happen during a manual test, I'd want to know.  Other than that, I have to extend *some* trust to AWS that they will send me the right stuff

reactor/discovery/curate_resource_handler.py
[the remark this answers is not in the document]
killthrush: +1 on `deep_get`, I will try that.  Re: `OldImage` INSERTs have NewImage, REMOVEs have OldImage, and MODIFY has both.  I think that warrants a comment - there is no ordering implied here despite the appearance

reactor/discovery/curate_resource_handler.py
killthrush: Good catch, I found that and fixed it in my tests but neglected to fix it here

reactor/discovery/facade/resource_tracking_table.py
killthrush: That's a great question, and this is looking weird to me now that you mention it.  I'll see if I can simplify

reactor/discovery/receive_events_handler.py
killthrush: As above, this is something under AWS' control.  If it fails I want the failure to be visible

reactor/discovery/receive_events_handler.py
killthrush: I'll try that

test/unit/discovery/test_cloudtrail_event_consumer.py
killthrush: Originally, I was also passing environments to CloudTrailEventConsumer, but that was refactored away.  It is still needed in `normalize` so this mock is still needed.  That said, CloudTrailEvent can probably be refactored too at  some point.  I don't think it needs to have such a strong dependency

reactor/discovery/curate_resource_handler.py
Tankanow: Fair enough. You've commented it nicely in the docstring. No further questions your honor.

reactor/discovery/receive_events_handler.py
Tankanow: Fair enough. You've commented it nicely in the docstring. No further questions your honor.

test/unit/discovery/test_cloudtrail_event_consumer.py
Tankanow: OK. Good enough for now. 👍

reactor/discovery/receive_events_handler.py
killthrush: Looks like it works!  The composition order still matters, with the end result being that the catch-all may or may not apply to code running inside the `accountid_aware` decorator

silvexis:
So I missed this before or wasn't on the PR but why have some of the dynamodb abstractions escaped from database\models\dynamodb\ out into this new folder discovery\facade?

Tankanow:
Hey @silvexis, the `discovery/facade` code is not new to this PR. The general idea is to have feature-specific abstractions close to the code that uses them rather than the more general `database/models/dynamodb`. I like the feature-specific direction, though I don't like being in limbo between both worlds. =)

reactor/discovery/facade/resource_tracking_table.py
Tankanow: I didn't see any changes to this code. Was this not fixable?

silvexis:
Hmmmm...what happens when I need to access one of these tables outside of this feature? Asking for a friend. I can be swayed either way, but would like us to discuss/socialize changes that break with convention.

reactor/discovery/facade/resource_tracking_table.py
killthrush: This was changed - see line 93

killthrush: It doesn't make sense to do `return response.get('Items')` because I want either a dict or None to be returned.  I just thought the use of `records = []` was weird

killthrush: actually, LOL, I just realized that I should be using get-item instead of query.  duh, that would be much better

PR #22, opened 2017-10-20
REACTOR-247 - Pick and connect to a code coverage tool for CLI and Re…

tiyberius:
…actor

### Description

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


sonar-project.properties
```diff
@@ -0,0 +1,32 @@
+# Required metadata
```
killthrush: remove project version on line 4?

tiyberius: I can take it out, but I was leaving it as a reminder that when we do have versioning figured out, we should populate it. It won't be populated here, it will be passed as an arg when sonar-scanner is run

PR #24, opened 2017-10-23
Hotfix/reactor 261

killthrush:
### Description
The occasional errors seen after the 10/17 prod deploy (RequestId:
057e950c-fea6-4f62-979e-adb40570155f Process exited before completing
request) are caused by the lambda running out of memory after being
called a number of times.

This was caused by having a max size of the tracking table's read
function set way too high.  It was assumed that the items would be
<= 1KB, but when in reality it was common to have items around 3KB.
Additionally, we neglected to leave memory for other things the
lambda might need to do.

Fix the bug by reducing the cache size by 75%; also add better debug
logging in order to keep tabs on cache usage going forward.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/utils/utils.py
```diff
@@ -33,6 +33,60 @@ def create_random_string(size=6, chars=None):
     return ''.join(random.SystemRandom().choice(chars) for _ in range(size))
```
killthrush: helpers for analyzing bytes and caching.  The first one I wrote, the second one I found on SO

requirements-dev.txt
```diff
@@ -14,8 +14,8 @@ pytest==3.1.3
 pytest-cov==2.5.1
```
killthrush: unrelated - pin these new deps I noticed

PR #25, opened 2017-10-24
REACTOR-261 - Remove read cache from tracking table due to memory issues

killthrush:
### Description
The initial REACTOR-261 fix didn't really show any improvements, but it did demonstrate how bad the cache hit rate was.  After concluding that 1) the cache is definitely taking up more memory than the sum of its stored elements, 2) that the cache is significant in size, and 3) that the hit rate was terrible, I decided to remove it outright.  Proved on staging that this ultimately fixed the issue.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


imcgaunn:
Since the hit rate is so bad, and the memory issues were so mysterious, I think these changes are uncontroversial. Very good debugging work and useful discovery of a memory profiler we can use with lambda.

PR #50, opened 2018-01-04
Write events to env-id folders

silvexis:
In addition fixes:
*  incorrect event location for athena
* over-abstraction of reactor bucket name
* Removes unnecessary need for GET/LIST permissions on CoreSync destination bucket


reactor/aws/cloudtrail.py
```diff
@@ -61,10 +62,10 @@ def write(self, env_id, data):
         Returns:
             The key used to store the given blob
         """
-        now = datetime.now(timezone.utc)
-        key = self.__generate_key(env_id, now)
         zipped_data = encode_and_compress(data)
         buffer = io.BytesIO(zipped_data)
+        salt = f"{sys.getsizeof(data)}-{sys.getsizeof(zipped_data)}"
```
tiyberius: should this go inside `__generate_key()`?

tox.ini
```diff
@@ -44,8 +44,7 @@ deps = -rrequirements-dev.txt
 ;setenv = PYTHONPATH = {toxinidir}/reactor
 commands =
     flake8                                                                   \
-        --output-file=.flake8                                                \
-        --format=codeclimate
```
tiyberius: thank you

silvexis: lol

reactor/aws/cloudtrail.py
silvexis: I thought about it but it's rather specific and I didn't like the idea of referencing two huge data structures (which data and zipped_data are potentially) outside of the main function...superstition paranoia

PR #54, opened 2018-01-15
REACTOR-338 - Update Reactor API to be more RESTful

tiyberius:
A note to make review of the Swagger document easier to read: you can take the `swagger.yml` file and throw it into Swagger's online editor, https://editor.swagger.io/ to get a visual representation of it instead of reading the yaml


lambda_handlers.py
```diff
@@ -63,10 +63,10 @@ def system__coresync_activate(event, context, namespace, configuration, client_d
 # ---------------------------------------------------------------------------------------------
 @lambda_handler()
 def configure__account__check(event, context, namespace, configuration, client_details):
```
killthrush: ah, ok, so the account ID is now part of the URL?

lambda_handlers.py
```diff
@@ -171,13 +166,10 @@ def resources__curate(event, context, namespace, configuration, client_details):
...
 def resources__get(event, context, namespace, configuration, client_details):
-    event_body = event.get('body')
     try:
-        event_map = json.loads(event_body)
-    except json.errors.JSONDecodeError:
-        logger.error('resources__get | Invalid JSON body received')
-        return generate_response(400, json.dumps({'error': 'Invalid request'}))
-    account_id = event_map.get('account_id')
+        account_id = event['pathParameters']['id']
+    except KeyError:
```
killthrush: do you know offhand if `generate_response` logs anything?  Not really a requested change, but eventually if I was returning > 200 status codes I'd want to log a warning or error

serverless.yml
```diff
@@ -158,79 +150,89 @@ functions:
...
-  query:
-    handler: lambda_handlers.simplequery__query
+  accounts-connect:
+    handler: lambda_handlers.configure__connect_account
     events:
       - http:
-          path: query
-          method: get
+          path: accounts
+          method: post
```
killthrush: typo on line 195 - `acccounts`

serverless.yml
```diff
@@ -158,79 +150,89 @@ functions:
...
+    handler: lambda_handlers.configure__disconnect_account
     events:
-      - sns: cz-reactor-${self:provider.stage}-system-setup-finalizer
-  system-coresync:
-    handler: lambda_handlers.system__coresync_activate
+      - http:
+          path: accounts/{id}
+          method: delete
+          private: true
+          request:
```
killthrush: nice!  I wasn't expecting path params to be this easy

swagger.yml
```diff
@@ -0,0 +1,800 @@
+swagger: "2.0"
```
killthrush: what's the process for maintaining swagger yaml?  done by hand?

lambda_handlers.py
tiyberius: that is correct. I pulled it out of the request and made it a path parameter in the URL

lambda_handlers.py
tiyberius: It does not currently log anything. However, I definitely think it should. We need to get better at errors and alerting, currently we don't have have much set up that will tell us if there are errors being thrown

serverless.yml
tiyberius: ooh great catch!

serverless.yml
tiyberius: right! me too. It was a pleasant surprise

silvexis: holy crap that's easy

reactor/features/simplequery/initialize.py
```diff
@@ -34,7 +34,7 @@ def initialize(namespace):
     logger.info('Starting Athena DB and Events Table Initialization')
 
     try:
-        bucket = get_reactor_bucket_name()
+        bucket = get_reactor_bucket_url()
         logger.info('Initializing Athena. Query results will be stored in {}.'.format(bucket))
         conn = _connection(bucket)
```
silvexis: so before this was just a bucket name and now it's a bucket URL, but I didn't see any changes to _connection so either we _had_ a bug or we now _have_ a bug

lambda_handlers.py
Tankanow: this repeated `KeyError` checking may mean there is use in `pyfaaster` for path parameters.

reactor/aws/tools.py
```diff
@@ -481,6 +481,19 @@ def get_reactor_bucket_name():
     return os.environ['CONFIG']
 
 
+def get_reactor_bucket_url():
+    reactor_bucket_name = get_reactor_bucket_name()
+    return get_query_results_bucket_url(reactor_bucket_name)
+
+
+def get_query_results_bucket_url(bucket_name):
```
Tankanow: this method should be called something like `bucket_name_to_url`

serverless.yml
```diff
@@ -158,79 +150,89 @@ functions:
               - CzReactorResources
               - StreamArn
   # APIs
-  configure-connect-account:
-    handler: lambda_handlers.configure__connect_account
+  reactor-setup:
```
Tankanow: The `/reactor` path is a little weird to me b/c the system itself is the reactor. In the past I have prefixed system level endpoints with `/system` as we did before ... though your relative paths are much nicer now. I'm not saying you have to change this. Just that I'm having a minor allergic reaction.

silvexis: yeah i second this, I had the same allergic reaction. At first it was just a little reaction but now half of my face is swollen and I might need an epi pen. Can you change it to `/system` or make a strong argument for keeping it `/reactor`?

PR #59, opened 2018-01-30
REACTOR-346

silvexis:
- Fix bug that was regenerating env id
- Add env ID to account API responses
- Fix and update docs


reactor/features/accounts/connect.py
```diff
@@ -42,10 +42,19 @@ def connect_account(account_details, cfg, client_details, reactor_account_id):
     logger.info(f'connect_account | Attempting to connect account {account_id}')
     logger.debug(f'connect_account | {account_details}')
     slack = SlackWeb(cfg.get('slack_webhook_url'))
+
+    # check if we have registered this account before, if so do not generate new env-id
+    env_dbh = dynamodb.CZReactorEnvironments()
+    try:
+        environment = env_dbh.get_from_account_id(account_id)
+        environment_id = environment['environment_id']
+    except dynamodb.DataNotFound:
```
killthrush: I have mixed feelings about making a "safe" operation throw exceptions, but OTOH python loves throwing exceptions =)

reactor/features/accounts/connect.py
```diff
@@ -55,20 +64,25 @@ def connect_account(account_details, cfg, client_details, reactor_account_id):
         if not subscribed:
             raise RegistrationFailure("Could not subscribe to target SNS topic")
 
-        success = _save_environment_details(account_details, client_details, verify_details)
-        if not success:
+        env_data = _save_environment_details(account_details, client_details, verify_details, environment_id)
```
killthrush: much better var name

reactor/features/accounts/connect.py
```diff
@@ -80,10 +94,10 @@ def connect_account(account_details, cfg, client_details, reactor_account_id):
                                       "text": str(error)},
                                      {"title": 'Client:',
                                       "text": client_details}])
-        return generate_response(502, json_verify_details)
+        return generate_response(502, json.dumps(verify_details, cls=jsontools.EnumEncoder))
 
 
-def _save_environment_details(account_details, client_details, verify_details):
+def _save_environment_details(account_details, client_details, verify_details, environment_id=None):
```
killthrush: how come the `environment_id` is not a required arg?

reactor/features/accounts/connect.py
Tankanow: Yes. I think we should reserve our own exceptions for truly out-of-the-ordinary cases. For example, if `get_from_account_id` returning `{}` is "normal", it should do so and this code can use `.get('environment_id')` instead. As a general rule of 👍, we may want to adopt a bit of Erlang's "let it crash" ... and only catch exceptions if the code can recover.

docs/connect-account.rst
```diff
@@ -9,27 +9,28 @@ This document will walk you through the process of connecting an AWS account to
...
 Here are the steps required to configure your account for CloudZero:
 
-- Create an S3 bucket for CloudZero
 - Setup an SNS topic for CloudTrail
 - Create an SQS queue and subscribe it to the CloudTrail SNS topic
 - Setup a CloudTrail
 
   - Events will be forwarded to new SNS topic
-  - Events will be stored in new S3 bucket
+  - Events will be stored in your CloudTrail S3 bucket
```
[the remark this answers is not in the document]
silvexis: we do have a reactor bucket, it’s however managed and configured by the reactor and not part of account setup

reactor/database/models/dynamodb/cz_reactor_environments.py
```diff
@@ -114,8 +117,11 @@ def get_from_environment_id(self, environment_id):
             return self.table.query(IndexName='EnvIdGSI',  # THIS ID IS SET IN SERVERLESS.YML
                                     KeyConditionExpression=Key('environment_id').eq(environment_id)).get('Items', [])[0]
         except botoerr.ClientError as err:
-            logger.error('ClientError while trying to get account_id {}: {}'.format(environment_id, err))
-            raise
+            if 'ResourceNotFoundException' in str(err):
+                raise common.DataNotFound('Environment id does not exist')
+            else:
+                logger.error('ClientError while trying to get account_id {}: {}'.format(environment_id, err))
```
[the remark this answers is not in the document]
silvexis: I don’t actually want a stack trace, throwing an exception is the desired action

reactor/features/accounts/connect.py
silvexis: I don’t want it to crash, I want to catch the exception because the function has an lru cache wrapper. Any return value other than exception would be cached.

Also embrace exceptions and control flow, this is python

reactor/features/accounts/connect.py
silvexis: If I don’t have one, it creates one for me

reactor/database/models/dynamodb/cz_reactor_environments.py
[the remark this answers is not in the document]
killthrush: Should still fix the message though

reactor/features/accounts/connect.py
killthrush: Good point about the LRU cache - that makes sense if you don't want to cache `None`

reactor/features/accounts/connect.py
killthrush: Ah, just wasn't apparent by looking at the code.

docs/api.rst
```diff
@@ -0,0 +1,5 @@
+CloudZero Reactor API
+=====================
+The `CloudZero Reactor API`_ is documented on SwaggerHub
+
+.. _CloudZero Reactor API: https://app.swaggerhub.com/apis/cloudzero/cloudzero-reactor-api/
```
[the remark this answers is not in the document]
silvexis: it can update it automatically but i haven't wired it up yet.

PR #64, opened 2018-02-27
REACTOR-349 - Fix blank event_actor_resource_ids

tiyberius:
### Description
In the Toast data, we found an intolerably large percentage of AssumeRole events not having a value for the event_actor_resource_id field. This applies to other customers as well, but needed to be addressed

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/aws/cloudtrail.py
```diff
@@ -382,12 +380,80 @@ def traverse_map(obj, map):
...
+    "principalId": "BARAJ75YVPL2FEZ5ZOCG1" }
+    >>> get_resource_id_from_user_identity(user_identity_field, region='us-east-1')
+    'czrn:aws::us-east-1:::::awsaccount:111111111111'
+
+    >>> get_resource_id_from_user_identity(None)
     """
-    return full_servicename.split('.')[0]
+    try:
+        resource_id = user_identity_field['arn']
+    except (AttributeError, KeyError, TypeError):
```
imcgaunn: how would a `TypeError` happen here?

reactor/aws/cloudtrail.py
```diff
@@ -382,12 +380,80 @@ def traverse_map(obj, map):
...
+            resource_id = construct_resource_id_from_user_identity_by_type(user_identity_field, region, service)
+        except IncompleteCloudtrailDataException as error:
+            logger.error(error)
+            resource_id = None
+    return resource_id
+
+
+def construct_resource_id_from_user_identity_by_type(user_identity_field, region, service):
+    try:
+        type = user_identity_field['type']
```
imcgaunn: `type` is a python reserved word, so I would recommend calling this `identity_type`.

imcgaunn: I also have the same question as above regarding `TypeError`

reactor/aws/cloudtrail.py
```diff
@@ -382,12 +380,80 @@ def traverse_map(obj, map):
...
+        arn = construct_resource_id_from_aws_account_type(user_identity_field, region, service)
+    elif type == 'AWSService':
+        arn = construct_czrn_from_aws_service_type(user_identity_field, region)
+    else:
+        raise IncompleteCloudtrailDataException('User Identity Field does not have supported "type" field')
+    return arn
+
+
+def construct_resource_id_from_aws_account_type(user_identity_field, region, service):
+    try:
```
imcgaunn: you could replace these `try / except AttributeError` blocks with `user_identity_field.get(<field>, '')`. I think that is a more concise pattern for assigning a default value if a field is missing.

reactor/aws/cloudtrail.py
```diff
@@ -382,12 +380,80 @@ def traverse_map(obj, map):
...
+    resource_id = get_resource_id_from_principal_id(principal_id)
+    resource_type, resource_service = determine_resource_type_and_service(resource_id)
+    service = resource_service or service
+    if resource_type is None:
+        resource_type = 'unknown'
+        if resource_id is None:
+            resource_type = 'awsaccount'
+            resource_id = account_id
+            account_id = None
+        resource_id = czrn.create(service, region, resource_id, account_id, resource_type=resource_type)
```
imcgaunn: perhaps not a concern for this PR, but it seems like there is some asymmetry between the `czrn.create` function and the `construct_arn` function regarding positional/keyword args. I would expect the way that I pass in a `resource_type` to be the same in both cases.

reactor/aws/cloudtrail.py
```diff
@@ -466,6 +512,14 @@ def get_arn_from_request_parameters(request_parameters):
             pass
 
 
+def get_service_from_full_servicename(full_servicename):
```
imcgaunn: maybe replace `service name` with `service uri`? what you are trying to return is the service name from amazon's url, right?

reactor/aws/decorators.py
```diff
@@ -24,3 +25,17 @@ def inner(*args, **kwargs):
         return inner
 
     return wrapper
+
+
+def convert_nonetype_args_to_empty_strings(func_to_decorate):
```
imcgaunn: `nonetype` in this case would appear to be `falsy` types. This will coerce all falsy args to empty strings.

reactor/utils/czrn.py
```diff
@@ -30,67 +32,135 @@ def create(input_string, service, region, account_id=None, network_id=None, subn
...
-    subnet_id = subnet_id or ""
-    interface_id = interface_id or ""
-    provider = "aws"
+    service = service or ''
+    region = region or ''
+    resource = resource or ''
+    account_id = account_id or ''
+    network_id = network_id or ''
+    subnet_id = subnet_id or ''
+    interface_id = interface_id or ''
```
imcgaunn: not even necessary with that decorator applied!

reactor/utils/czrn.py
```diff
@@ -30,67 +32,135 @@ def create(input_string, service, region, account_id=None, network_id=None, subn
...
+    resource_type = 'unknown'
+    try:
+        ip = ipaddress.ip_interface(resource)
+        resource_type = f'ipv{ip.version}_address'
     except ValueError:
-        if is_valid_hostname(input_string):
-            resource = input_string
+        if is_valid_hostname(resource):
             resource_type = "host"
+        elif re.match("[0-9a-f]{2}([-:])[0-9a-f]{2}(\\1[0-9a-f]{2}){4}$", resource.lower()):
```
imcgaunn: Would  be nice if this macaddr regular expression were a named constant somewhere else.

reactor/utils/czrn.py
```diff
@@ -30,67 +32,135 @@ def create(input_string, service, region, account_id=None, network_id=None, subn
...
+def determine_provider(resource):
+    """
+    >>> determine_provider('10.0.1.45')
+    'aws'
+    >>> determine_provider('54.0.1.45')
+    'internet'
+    """
+    try:
+        ip = ipaddress.ip_interface(resource)
+        provider = ip.is_private and 'aws' or 'internet'
```
imcgaunn: this works, but is less clear than `provider = 'aws' if ip.is_private else 'internet'`

reactor/utils/czrn.py
```diff
@@ -30,67 +32,135 @@ def create(input_string, service, region, account_id=None, network_id=None, subn
...
+    except (ValueError, TypeError):
+        provider = 'aws'
+    return provider
 
 
 def is_valid_hostname(hostname):
+    """
+    >>> is_valid_hostname('cloudzero.com')
+    True
+    # todo shouldn't this evaluate to False?
```
imcgaunn: I would make this todo capital. I think that it shows up better in IDEs when its stylized that way.

test/data/sample_normalized_event_data.py
```diff
@@ -18,7 +18,7 @@
     "environment_type": "AWS",
     "cloud_account_id": "111111111111",
     "event_actor_resource_id": "arn:aws:ec2:us-east-1:444444444444:instance/i-0e90dc30b5bbe92cb",
-    "event_source_resource_id": "czrn:internet:AwsApiCall:us-east-1:111111111111::::ipv4_address:111.111.111.111/32",
+    "event_source_resource_id": "czrn:internet::us-east-1:111111111111::::ipv4_address:111.111.111.111/32",
```
imcgaunn: are these replacements in the test data intentional?

reactor/aws/cloudtrail.py
tiyberius: If the user_identity_field is None, a TypeError is thrown

```>>> a = None
>>> a['arn']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not subscriptable```

File: reactor/aws/cloudtrail.py (hunk 2 above)
Comment (tiyberius) [reply reactor/aws/cloudtrail.py]: Ahhh great call! I'll change it

Comment (tiyberius) [reply reactor/aws/cloudtrail.py]: "Same question as above regarding `TypeError`", what do you mean by that?

File: reactor/aws/cloudtrail.py (hunk 3 above)
Comment (tiyberius) [reply reactor/aws/cloudtrail.py]: So that was my initial approach, but when `user_identity_field` is `None`, you get an `AttributeError`

```
>>> user_identity_field = None
>>> user_identity_field.get('accountId')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'get'
```

reactor/aws/cloudtrail.py
tiyberius: Yes, definitely some asymmetry. I wrote `construct_arn` and Erik wrote `czrn.create()`. I like the `czrn.create()` style better, I would like to fix it in the future

reactor/utils/czrn.py
tiyberius: Ah! That is what I thought originally too! BUT if someone does not pass an argument explicitly, it does not get passed to the decorator. If you comment out `region = region or ''` and run the doctests, you'll see that they fail

reactor/utils/czrn.py
tiyberius: I agree. I was thinking of pulling this out into its own thing. I am going to do that, it will be in the next PR update

reactor/utils/czrn.py
tiyberius: YES. Thank you! I had the same thought but didn't change it. Will be changed in the next PR update

test/data/sample_normalized_event_data.py
tiyberius: Yup! All intentional. It was changed it in a previous ticket and was inaccurate

reactor/utils/czrn.py
tiyberius: Roger that, will do 👍

reactor/aws/decorators.py
tiyberius: Ahhh you're right. I actually had an explicit check for None but changed it, I will change it back

reactor/aws/cloudtrail.py
tiyberius: So I picked `service name` because it showed up in the AWS docs. I was reading this https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-record-contents.html and if you scroll to where they cover the `eventSource` field, you'll see that they say "The service that the request was made to. This name is typically a short form of the _service name_ without spaces plus .amazonaws.com. For example: ..." So when they said "short form of the service name", I used "full service name" to describe the opposite of that. Not sure I explained that very well, does that make sense?

imcgaunn: maybe then something like `get_shortname_from_full_servicename` would be better? Although The function's name makes it sound like it is returning an object called a `service` from a string representing a name. This function is so simple though that it might be worth just using inline.

reactor/utils/czrn.py
imcgaunn: that's lame. what is the decorator for, then?

reactor/aws/cloudtrail.py
tiyberius: Ah, yes that makes it clearer. It is changed

reactor/utils/czrn.py
tiyberius: removed :)

PR #65, opened 2018-03-08
Local Flow

silvexis:
Flow Logs for local reactor vpc's


reactor/cz_nef.py
```diff
@@ -79,37 +78,41 @@ def normalize_vpc_flow_log_event(environment_id, event_id, flowlog_event):
...
+    new_event['event_region'] = eni_details['cloud_region']
     new_event['event_version'] = flowlog_event['version']
 
     data = {"protocol": flowlog_event['protocol'],
             "packets": flowlog_event['packets'],
             "bytes": flowlog_event['bytes'],
             "action": flowlog_event['action'],
             "log-status": flowlog_event['log_status']}
-    new_event['event_data'] = json.dumps(data)
+    new_event['event_data'] = data
```
tiyberius: Ha, I believe I was there for this change

reactor/cz_nef.py
```diff
@@ -79,37 +78,41 @@ def normalize_vpc_flow_log_event(environment_id, event_id, flowlog_event):
...
-                                           eni_details['cloud_resource_region'],
+                                           eni_details['cloud_region'],
                                            flowlog_event['dstaddr'],
                                            flowlog_event['account_id'],
                                            network_id=eni_details['cloud_vpc_id'],
                                            interface_id=flowlog_event['interface_id'])
     new_event['event_target_resource_id'] = event_target_resource_id
 
-    new_event['event_datetime'] = ciso8601.parse_datetime(flowlog_event['start'])
+    new_event['event_datetime'] = datetime.fromtimestamp(int(flowlog_event['start']), timezone.utc).isoformat()
```
tiyberius: Was this not working or you just like the way this looks better?

reactor/features/resources/get.py
```diff
@@ -33,17 +33,39 @@ def get_resources(cloud_account_id):
         resources_dbh = dynamodb.CZReactorResources()
 
         ec2_count = get_ec2_resources(aws, resources_dbh, env_settings, cloud_account_id, region)
+        eni_count = get_eni_resources(aws, resources_dbh, env_settings, cloud_account_id, region)
```
tiyberius: Hmmm this will have to be refactored at some point. Not right now but when we add resource types, it will get messy

reactor/cz_nef.py
silvexis: it wasn't actually working, the input was a timestamp, not a iso string

tiyberius: 👍

PR #67, opened 2018-03-25
REACTOR-355 - Generalized Enrichment

silvexis:
REACTOR-355 | Configuration driven enrichment of resources allowing us to extract information on all resources without having to write plugins for each one


reactor/aws/tools.py
```diff
@@ -156,25 +152,47 @@ def get_resource_type_from_arn(arn):
...
 
 
     >>> get_resource_type_from_arn('arn:partition:service:region:account-id:resourcetype/resource')
     'resourcetype'
 
+    >>> get_resource_type_from_arn('arn:aws:iam::account-id:role/service-role/config-role-us-east-1')
+    'service-role'
+
     """
+    split = arn.split(':')
```
imcgaunn: split is a confusing name for a number of reasons, but mostly because it's the same name as the `split` method.

reactor/constants.py
```diff
@@ -34,12 +34,12 @@ def __str__(self):
...
 EVENT_SOURCE_CLOUDTRAIL = 'cloudtrail'
-RESOURCE_DISPOSITION_NEW = 'new'                        # The resource ID was just discovered
-RESOURCE_DISPOSITION_REFRESHED = 'refreshed'            # The resource ID just expired and was refreshed
-RESOURCE_DISPOSITION_IDENTIFIED = 'identified'          # We've done a checked and confirmed the resource
-RESOURCE_DISPOSITION_PENDING = 'pending'                # We're sure this is a resource, but confirmation is needed
-RESOURCE_DISPOSITION_NOT_SUPPORTED = 'not_supported'    # We tried to identify the resource but couldn't
-RESOURCE_DISPOSITION_ERROR = 'error'                    # unknown error attempting to process this resource
+RESOURCE_DISPOSITION_NEW = 'new'                          # The resource ID was just discovered
+RESOURCE_DISPOSITION_REFRESHED = 'refreshed'              # The resource ID just expired and was refreshed
+RESOURCE_DISPOSITION_IDENTIFIED = 'identified'            # We've identified the resource
```
imcgaunn: I like "we've done a checked" better. It's like Mario is telling us about the resource.

reactor/database/models/dynamodb/cz_reactor_discovery_map.py
```diff
@@ -0,0 +1,128 @@
...
+        'AttributeDefinitions': [{'AttributeName': 'service_type', 'AttributeType': 'S'},
+                                 {'AttributeName': 'resource_type', 'AttributeType': 'N'}],
+        'ProvisionedThroughput': {'ReadCapacityUnits': 1, 'WriteCapacityUnits': 1}
+    }
+
+    def __init__(self, namespace: str = None, aws: Provider = None):
+        super().__init__(self._TABLE_BASENAME, namespace, aws)
+
+    # @provisioned_throughput_retry(max_attempts=2)
+    def set_map(self, service_type, resource_type, metadata=None, enrich_functions=None):
```
imcgaunn: I would recommend adding something to these method names to make their function more clear. Looking at these. I get a little confused because `set` and `map` can both be verbs. Maybe expand `map` to `service_resource_map` so it's clear that it's a noun.

reactor/database/models/dynamodb/cz_reactor_resources.py
```diff
@@ -15,9 +18,11 @@
 from reactor.aws.tools import get_service_type_from_arn, get_resource_type_from_arn, get_object_name_from_arn, \
     get_region_from_arn
 from reactor.database.models.dynamodb.common import CZReactorModel
-from reactor.features.resources.plugins.base import create_resource_record
 from reactor.database.models.dynamodb import common
 from reactor.constants import DYNAMODB_RESOURCE_TABLE_GSI_NAME
+from reactor.utils.czrn import construct_arn
```
imcgaunn: I think you were using `from` imports for consistency here, which is noble, but our styleguide prefers that these not exist .

reactor/database/models/dynamodb/cz_reactor_resources.py
```diff
@@ -151,3 +156,69 @@ def query_by_resource_name_with_filter(self, object_name, filter_expression):
             else:
                 logger.error('ClientError while trying to get environment_id {}: {}'.format(object_name, err))
                 raise
+
+
+def create_resource_record(service_type, resource_name, resource_type, region, cloud_account_id, environment_id,
```
imcgaunn: these parameters should probably not be explicitly named in the args list. there are too many! maybe a namedtuple / struct type would be useful here?

imcgaunn: or use **kwargs.

reactor/features/resources/curate.py
```diff
@@ -22,7 +23,8 @@
 import reactor.constants
 import reactor.utils as utils
 from reactor.aws.provider import Provider
-from reactor.aws.tools import get_service_type_from_arn, get_reactor_bucket_name, get_object_name_from_arn
+from reactor.aws.tools import get_service_type_from_arn, get_reactor_bucket_name, get_object_name_from_arn, \
```
imcgaunn: seems like we should have an `ARN` library at this point. Lots of functions that operate specificly on ARNs.

reactor/features/resources/curate.py
```diff
@@ -82,27 +86,33 @@ def curate(records, reactor_account_id, namespace):
...
                 logger.debug(f"get_aws_account_details_for_env_id: {environment_id} "
                              f"cache: {get_aws_account_details_for_env_id_cached.cache_info()}")
 
                 if not aws_account_details:
                     logger.warning(f'Could not load aws_account_details for environment {environment_id}')
                     return
-                jobs.append(gevent.spawn(process_new_resource, resource_id, timestamp, service_type, environment_id,
-                                         cz_metadata, aws_account_details, dbh))
+
+                # TODO Make update_map take a count and sqoosh identical service/resource types
```
imcgaunn: `sqoosh` -> `squoosh`

reactor/features/resources/curate.py
```diff
@@ -82,27 +86,33 @@ def curate(records, reactor_account_id, namespace):
...
+                try:
+                    dbh_discovery_map.update_map(service_type, resource_type)
+                except Exception as err:
+                    logger.exception(err)
+                    logger.error(f"Error trying to map {service_type}.{resource_type} ({resource_id})")
+
+                jobs.append(gevent.spawn(process_new_resource, resource_id, timestamp, service_type, resource_type,
```
imcgaunn: should refactor this to use python's `async/await` event loop instead of `gevent`.

reactor/features/resources/plugins/no_plugin.py
```diff
@@ -24,19 +26,48 @@ def enrich_resource(event):
...
+
     try:
-        processor = Resource(resource_id=event['resource_id'],
-                             timestamp=event['timestamp'],
-                             environment_id=event['environment_id'])
-        logger.debug(f'no_plugin.enrich_resource | Received {processor.resource_id}')
-        processor.reconcile({"message": "No Plugin Available"}, reactor.constants.RESOURCE_DISPOSITION_NOT_SUPPORTED)
-        logger.info(f"no_plugin.enrich_resource | Processed {processor.resource_id}:{processor.timestamp}")
+        aws_account_details = event['aws_account_details']
+        resource = Resource(resource_id=event['resource_id'],
```
imcgaunn: a little confused here. what's the difference between a `Resource` object and a `resource_record` as generated by `create_resource_record`.

reactor/features/resources/curate.py
silvexis: I tried and I think I ran into something that told me async/await didn't work in Lambda...I need to research this again.

reactor/database/models/dynamodb/cz_reactor_discovery_map.py
imcgaunn: When I commented, didn't realize this was a class method. I think the fact that it's a method will be sufficient.

reactor/features/resources/plugins/ec2.py
```diff
@@ -71,55 +79,69 @@ def enrich_resource(event):
...
-        if 'InvalidInstanceID.NotFound' in str(error):
+        resource_status = reactor.constants.RESOURCE_STATUS_UPTODATE
+
+        create_eni_resources_from_ec2_instance(instance, resource)
+        if instance.vpc:
+            create_vpc_resources_from_ec2_instance(instance, resource)
+
+    except botocore.exceptions.ClientError as error:
+        if 'NotFound' in str(error):
             logger.debug(f"Instance {resource.resource_id} went away {error}")
```
killthrush: should this be `info`?

reactor/features/resources/plugins/ec2.py
```diff
@@ -215,57 +241,73 @@ def create_vpc_flow_log(vpc_id, logs_provider, ec2_provider):
...
+        logger.exception(error)
         if "UnauthorizedOperation" in str(error):
-            logger.warning(f'Reactor role {ec2_provider.role_arn} '
+            logger.warning(f'Reactor role {logs_provider.role_arn} '
                            f'does not have permission to create vpc flow logs for {vpc_id} ({error})')
+            return None
         else:
-            # If this failed it's almost certainly because the flow log was already setup, which is fine
-            logger.info(f'Flow logs already exists for {vpc_id} ({error})')
+            logger.warning(f'Flow logs not created for {vpc_id} ({error})')
```
killthrush: add `logger.exception` here?

reactor/features/resources/plugins/no_plugin.py
```diff
@@ -24,19 +27,65 @@ def enrich_resource(event):
...
+    service_type = get_service_type_from_arn(event['resource_id'])
+    resource_type = get_resource_type_from_arn(event['resource_id'])
+    object_name = get_object_name_from_arn(event['resource_id'])
+
+    # Detect if observed resource exists outside of the currently connected account
+    resource_cloud_account_id = get_account_id_from_arn(event['resource_id'])
+    connected_cloud_account_id = get_account_id_from_arn(aws_account_details['role_arn'])
+    if resource_cloud_account_id != connected_cloud_account_id:
+        disposition = reactor.constants.RESOURCE_DISPOSITION_IDENTIFIED
+        resource_status = reactor.constants.RESOURCE_STATUS_CROSS_ACCOUNT
```
killthrush: is `cross-account` really a status?  This seems like an immutable property of a resource

killthrush: Are we using it in a way to prevent unnecessary lookups?  kinda confused by this

reactor/features/resources/plugins/no_plugin.py
```diff
@@ -24,19 +27,65 @@ def enrich_resource(event):
...
+    else:
+        discovery_map = get_discovery_map_cached(service_type, resource_type)
+        logger.info(f"Discovery Cache: {get_discovery_map_cached.cache_info()}")
+
+        describe_function = discovery_map.get('enrich_functions', {}).get('describe')
+        if describe_function:
+            logger.info(f"Enriching {service_type}.{resource_type}.{object_name}")
+            provider = Provider(role_arn=aws_account_details['role_arn'],
+                                external_id=aws_account_details['external_id'],
+                                region=resource.region or 'us-east-1',
```
killthrush: if a resource region is not defined, can we really assume `us-east-1`?  I was under the impression if there's no region then it's meaningless

reactor/features/resources/plugins/no_plugin.py
```diff
@@ -24,19 +27,65 @@ def enrich_resource(event):
...
+    if resource_cloud_account_id != connected_cloud_account_id:
+        disposition = reactor.constants.RESOURCE_DISPOSITION_IDENTIFIED
+        resource_status = reactor.constants.RESOURCE_STATUS_CROSS_ACCOUNT
+        data = {"message": "Observed resource exists outside account",
+                "true_cloud_account_id": resource_cloud_account_id}
+    else:
+        discovery_map = get_discovery_map_cached(service_type, resource_type)
+        logger.info(f"Discovery Cache: {get_discovery_map_cached.cache_info()}")
+
+        describe_function = discovery_map.get('enrich_functions', {}).get('describe')
```
killthrush: based on what I've seen thus far, a purely data-driven approach to this problem is likely a fool's errand.  What does the update process look like for when we need to go find an enrich function by hand?  What if one does not exist?

reactor/features/resources/plugins/no_plugin.py
```diff
@@ -24,19 +27,65 @@ def enrich_resource(event):
...
+            parameter = {describe_function['parameter']: object_name}
+            boto3_function = getattr(provider.client, describe_function['function'])
+            try:
+                data = boto3_function(**parameter).get(describe_function['result'])
+                disposition = reactor.constants.RESOURCE_DISPOSITION_IDENTIFIED
+                if describe_function.get('state'):
+                    pass
+                else:
+                    resource_state = reactor.constants.RESOURCE_STATE_PRESENT
+            except ClientError as err:
```
killthrush: log `exception` here?

reactor/features/system.py
```diff
@@ -315,3 +316,40 @@ def status(account_id, configuration):
         logger.debug(configuration.json())
 
     return generate_response(status_code, json.dumps(reactor_details))
+
+
+def information(account_id, configuration):
+    """
+    Return PUBLIC information on this reactor. This function is NOT authenticated and is required to support
+    provisioning by users who do not have authenticated access to this Reactor
```
killthrush: what's the use case for this exactly?  Could you add to the docstring please?

reactor/utils/reasons.py
```diff
@@ -7,17 +7,19 @@
...
+        "Heavy gravity fluctuations",
         "Complete Transient Lockout averted",
         "CPU recalibration complete",
-        "disk or processor fire extinguished",
+        "Disk and/or processor fire extinguished",
         "Additional computrons made available",
-        "endothermal recalibration complete"
+        "Endothermic recalibration complete",
+        "Successfully reticulated splines",
+        "Additional pylons constructed"
```
killthrush: LOL

requirements-dev.txt
```diff
@@ -2,23 +2,20 @@
...
+docker==3.2.1
 flake8==3.5.0
 flake8-copyright==0.2.0
 flake8-codeclimate==0.1.3
 hypothesis==3.49.1
-moto==1.2.0
-placebo==0.8.1
-pycobertura==0.10.1
-pytest==3.4.2
+moto==1.3.1
```
killthrush: side note - I'm working on testing a new version of moto that purports to have fixed the `requests` lib issue.  I don't think it's been released yet

serverless.yml
```diff
@@ -62,19 +63,54 @@ provider:
...
+#        comparisonOperator: GreaterThanThreshold
+#        pattern: 'ERROR'
+#      - name: warningAlarm
+#        metric: logWarnings
+#        threshold: 5
+#        statistic: Minimum
+#        period: 60
+#        evaluationPeriods: 1
+#        comparisonOperator: GreaterThanThreshold
+#        pattern: 'WARNING'
```
killthrush: oh, OK, looks like you did disable this

serverless.yml
```diff
@@ -109,18 +145,18 @@ functions:
...
   ingest-event:
     handler: lambda_handlers.ingest_events
-    memorySize: 2048
+    memorySize: 1024
   ingest-vpc-flow-log-events:
     handler: lambda_handlers.ingest_vpc_flow_log_events
   process-aws-cloudtrail-event:
     handler: lambda_handlers.process_aws_cloudtrail_event
-    memorySize: 2048
+    memorySize: 1024
```
killthrush: think this will have a positive effect on cost?

serverless.yml
```diff
@@ -245,9 +289,38 @@ resources:
     # NOTE: When adding DynamoDB tables, ensure there is a DependsOn attribute
     # that lists the preceding (above) table.  This daisy chain will prevent
     # hitting CloudWatch alarm creation API limits.
+    CzReactorDiscoveryMap:
+      Type: 'AWS::DynamoDB::Table'
+      DeletionPolicy: Retain
+      Properties:
+        TableName: cz_reactor_discovery_map_${self:provider.stage}
+        SSESpecification:
+          SSEEnabled: true
```
killthrush: cool, you remembered to add this

reactor/database/models/dynamodb/cz_reactor_resources.py
silvexis: readability trumps everything. I see nothing wrong with from imports when it makes the code easier to read. I'm unsure how this developed into the point of view that "from imports are bad". There is no python style guide i'm aware of that makes that claim.

reactor/database/models/dynamodb/common.py
```diff
@@ -29,6 +28,10 @@ class CZReactorModel(object):
 
     def __init__(self, table_name: str, namespace: str = None, aws: Provider = None):
         self.namespace = namespace or os.environ.get('NAMESPACE')
+        if self.namespace is None:
+            logger.error('CZReactorModel Error, namespace must be set!')
+            return
```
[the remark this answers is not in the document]
silvexis: it's perfectly acceptable to raise an exception in a constructor in my book. However in this case, throwing an exception here will break all tests since the dynamodb connection objects are instantiated on load and namespace doesn't get set until after pytest loads everything, which is fine if we just short-circuit. The real purpose of short-circuiting this here is to just be a safety against a missing namespace when something real is attempted

reactor/database/models/dynamodb/cz_reactor_discovery_map.py
```diff
@@ -0,0 +1,136 @@
...
+                      {'AttributeName': 'resource_type', 'KeyType': 'RANGE'}],
+        'AttributeDefinitions': [{'AttributeName': 'service_type', 'AttributeType': 'S'},
+                                 {'AttributeName': 'resource_type', 'AttributeType': 'N'}],
+        'ProvisionedThroughput': {'ReadCapacityUnits': 1, 'WriteCapacityUnits': 1}
+    }
+
+    def __init__(self, namespace: str = None, aws: Provider = None):
+        super().__init__(self._TABLE_BASENAME, namespace, aws)
+
+    # @provisioned_throughput_retry(max_attempts=2)
```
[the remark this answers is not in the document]
silvexis: I actually need to put it back. The retry decorator can hide exceptions and makes debugging hard sometimes.

reactor/database/models/dynamodb/cz_reactor_discovery_map.py
```diff
@@ -0,0 +1,136 @@
...
+        Returns (dict):
+            The latest value of the updated record
+        """
+
+        count = decimal.Decimal(count)
+        result = self.table.update_item(Key={'service_type': service_type,
+                                             'resource_type': resource_type},
+                                        UpdateExpression=f'ADD observed_count :val1 '
+                                                         f'SET last_seen_datetime = :val2, '
+                                                         f'enrich_functions = if_not_exists(enrich_functions, :val3)',
```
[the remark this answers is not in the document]
silvexis: sure you can, that's the purpose of the set_map function

reactor/features/ingest.py
```diff
@@ -301,25 +301,40 @@ def write_events_to_bucket(events, message_source_environment_id, namespace, rea
...
-    logger.debug(resource_ids)
+        for resource in cloudtrail_event.normalized_event['event_resources']:
+            if is_arn_or_czrn(resource):
+                if get_resource_type_from_arn(resource) not in resource_types_to_ignore:
+
+                    # Slow the rate of unique resources discovered to 1 per <seen_window> seconds for this function
+                    # this is a simple, yet effective way of preventing certain resources that are seen repeatedly from
+                    # overwhelming the system
+                    if seen_resources.get(resource, 0) <= \
+                            cloudtrail_event.normalized_event['event_timestamp'] - seen_window:
```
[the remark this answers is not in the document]
silvexis: they contain resource_id:timestamp key value pairs. but ok i can rename to seen_resource_timestamps

reactor/features/ingest.py
```diff
@@ -301,25 +301,40 @@ def write_events_to_bucket(events, message_source_environment_id, namespace, rea
...
+                        resource_ids[resource] = cloudtrail_event.normalized_event['event_timestamp']
+                        seen_resources[resource] = cloudtrail_event.normalized_event['event_timestamp']
+                    else:
+                        skipped_res_count += 1
```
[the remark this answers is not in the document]
silvexis: I think the counts are kinda meaningless to be honest. So what if we have seen a specific resource type 5000 times in cloudtrail? I'm not sure it means anything useful. The only value I get out of it is a vague understanding that events are flowing as expected, which is useful, but the actual number isn't I think.

reactor/features/resources/plugins/base.py
```diff
@@ -142,7 +88,7 @@ def reconcile(self, data,
         Returns:
 
         """
-        logger.info(f"Reconciling {self.resource_id}")
+        logger.debug(f"Reconciling {self.resource_id}")
```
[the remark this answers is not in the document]
silvexis: it was ;)

reactor/features/resources/plugins/ec2.py
```diff
@@ -71,55 +79,69 @@ def enrich_resource(event):
...
+    except botocore.exceptions.ClientError as err:
         if "AccessDenied" in str(err):
             logger.error(f'Access Denied trying to enrich {resource.resource_id}. Was our access removed?')
             data = {}
-            resource_state = reactor.constants.RESOURCE_STATE_NO_ACCESS
-            resource_status = reactor.constants.RESOURCE_STATUS_UNKNOWN
+
+            disposition = reactor.constants.RESOURCE_DISPOSITION_PENDING
+            resource_state = reactor.constants.RESOURCE_STATE_UNKNOWN
+            resource_status = reactor.constants.RESOURCE_STATUS_NO_ACCESS
```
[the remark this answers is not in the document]
silvexis: we should probably rename resource_status to cz_resource_status

reactor/database/models/dynamodb/common.py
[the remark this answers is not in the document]
killthrush: yeah, the worst outcome is having an object that is invalid.  I agree with that ^^^

reactor/database/models/dynamodb/cz_reactor_discovery_map.py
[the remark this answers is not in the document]
killthrush: Oh, OK, I didn't zoom out far enough =)

reactor/database/models/dynamodb/cz_reactor_resources.py
killthrush: Readability is very important, yes, but something that's both readable and testable is even better.  That's why we agreed on the standard.  Pythonistas are not generally opinionated so I'm not going to use that as a guide per se - the prevailing guide is usually "whatever works for you"

reactor/database/models/dynamodb/cz_reactor_resources.py
```diff
@@ -151,3 +157,72 @@ def query_by_resource_name_with_filter(self, object_name, filter_expression):
...
+
+    Returns:
+        dict: A new resource record dictionary
+    """
+
+    if not timestamp:
+        timestamp = Decimal(str(datetime.now(tz=timezone.utc).timestamp())).quantize(Decimal(".00000"),
+                                                                                     rounding=ROUND_DOWN)
+    if not res_id:
+        res_id = construct_arn(service_type, region, cloud_account_id, resource_type, resource_name)
```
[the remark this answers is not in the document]
killthrush: Alright, then if you're not sure how to solve it, then this ^^^ is your solution.  Don't construct ARNs, just record them and build CZRNs instead.

reactor/features/ingest.py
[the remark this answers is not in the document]
killthrush: Please also rename `resource_ids` - same issue

reactor/features/ingest.py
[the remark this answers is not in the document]
killthrush: It's useful in that it's knowledge about AWS that's hidden.  It's useful for us in an academic/research sense I think, and may be useful for customers.  If your count is off by a few, no big deal - but if it's off by a big factor it may totally skew the results.  That's why I asked - if we erase this information, is there any way to recover it later?

reactor/database/models/dynamodb/cz_reactor_resources.py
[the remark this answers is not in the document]
killthrush: We should definitely discuss this - I see this as a pretty big flaw. But nothing that should block this PR =)

reactor/features/resources/plugins/ec2.py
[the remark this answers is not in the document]
silvexis: actually i'm thinking now that we should merge resource_status into disposition and make status go away.

requirements-dev.txt
imcgaunn: this would be very nice!!!

reactor/features/resources/get.py
```diff
@@ -23,23 +27,44 @@ def get_resources(cloud_account_id):
...
+                               ProjectionExpression="res_id,#cztimestamp",
+                               ExpressionAttributeNames={"#cztimestamp": "timestamp"})
+
+    purge_count = 0
+    for item in stale.get('Items'):
+        with res_dbh.table.batch_writer() as batch:
+            purge_count += 1
+            batch.delete_item(Key={"res_id": item['res_id'],
+                                   "timestamp": item['timestamp']})
+    logger.info(f"Purged {purge_count} of {len(stale.get('Items'))} stale resources")
```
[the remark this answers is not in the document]
killthrush: This is still important - a GET that causes database writes is an abomination =)  This should be a cron lambda

reactor/features/resources/plugins/ec2.py
```diff
@@ -2,43 +2,48 @@
...
 from reactor.database.models import dynamodb
+from reactor.database.models.dynamodb.cz_reactor_resources import create_resource_record
 
 logger = setup_logging('reactor')
 
 _SUPPORTED_RESOURCE_TYPES = ['instance', 'vpc', 'network-interface']
+resources_dbh = dynamodb.CZReactorResources()
 
 
 def enrich_resource(event):
```
killthrush: This module should be devoted solely to enriching EC2 instances.  Modules should be loaded for service type and resource type.  Otherwise this will quickly become a mess

reactor/features/resources/plugins/ec2.py
```diff
@@ -2,43 +2,48 @@
...
 from reactor.aws.provider import Provider
-from reactor.aws.tools import get_resource_type_from_arn
-from reactor.features.resources.plugins.base import Resource, create_resource_record
-import reactor.constants
 from reactor.database.models import dynamodb
+from reactor.database.models.dynamodb.cz_reactor_resources import create_resource_record
 
 logger = setup_logging('reactor')
 
 _SUPPORTED_RESOURCE_TYPES = ['instance', 'vpc', 'network-interface']
```
killthrush: Each of these should be a separate module that conforms to the design laid out in the wiki.

reactor/features/resources/plugins/ec2.py
```diff
@@ -49,161 +54,138 @@ def enrich_resource(event):
...
+                            region=event['region'],
                             service='ec2')
 
     try:
-        if resource.resource_type == 'instance':
-            data, resource_state, resource_status = process_instance(resource, ec2_provider)
-        elif resource.resource_type == 'vpc':
-            data, resource_state, resource_status = process_vpc(resource, ec2_provider)
+        if event['resource_type'] == 'instance':
+            data, resource_state = _process_instance(event, ec2_provider)
```
killthrush: Separate these into modules please

reactor/features/resources/plugins/ec2.py
```diff
@@ -49,161 +54,138 @@ def enrich_resource(event):
...
-                                          resource_state=instance.vpc.state,
-                                          resource_status=reactor.constants.RESOURCE_STATUS_PRESENT,
-                                          vpc_id=instance.vpc.id)
-    resources_dbh = dynamodb.CZReactorResources()
-    resources_dbh.write([vpc_resource])
-
-
-def create_eni_resources_from_ec2_instance(instance, resource):
+        # processed_count = resource.reconcile(data, disposition=disposition, resource_state=resource_state,
+        #                                      cz_metadata=cz_metadata)
```
killthrush: remove these lines?

reactor/features/resources/plugins/ec2.py
```diff
@@ -49,161 +54,138 @@ def enrich_resource(event):
...
         eni.load()
-        eni_resources.append(create_resource_record(service_type=resource.service_type, resource_name=eni.id,
-                                                    resource_type='network-interface', region=resource.region,
-                                                    cloud_account_id=resource.account_id,
-                                                    environment_id=resource.environment_id, resource_state=eni.status,
-                                                    resource_status=reactor.constants.RESOURCE_STATUS_PRESENT,
+        eni_resources.append(create_resource_record(service_type=service_type, resource_name=eni.id,
+                                                    resource_type='network-interface', region=region,
+                                                    cloud_account_id=cloud_account_id,
+                                                    environment_id=environment_id, resource_state=eni.status,
```
killthrush: Yeah, we need to talk about this - I don't think we're understanding each other on the design.

reactor/features/resources/plugins/ec2.py
```diff
@@ -49,161 +54,138 @@ def enrich_resource(event):
...
-        resource_status = reactor.constants.RESOURCE_STATUS_REMOVED
-
-    return data, resource_state, resource_status
+def create_vpc_resources_from_ec2_instance(instance, service_type, cloud_account_id, region, environment_id):
+    vpc_resource = create_resource_record(service_type=service_type, resource_name=instance.vpc.id,
+                                          resource_type='vpc', region=region,
+                                          cloud_account_id=cloud_account_id, environment_id=environment_id,
+                                          resource_state=instance.vpc.state,
+                                          vpc_id=instance.vpc.id)
+    resources_dbh.write([vpc_resource])
```
killthrush: Plugins should return relationships and `enrich` function should write them all.  This code should be as lean as possible

reactor/features/resources/plugins/ec2.py
```diff
@@ -2,43 +2,48 @@
...
     """
-    logger.debug(f"ec2.enrich_resource | Received {event['resource_id']}")
-    resource_type = get_resource_type_from_arn(event['resource_id'])
+    logger.info(f"Received for enrichment {event['resource_id']}")
+    data = None
+    resource_state = None
 
-    if resource_type not in _SUPPORTED_RESOURCE_TYPES:
-        logger.warning(f"Received a {resource_type} resource, but no resource processor available")
+    if event['resource_type'] not in _SUPPORTED_RESOURCE_TYPES:
```
killthrush: The input to enrich should be a resource record, not an event.

reactor/features/resources/plugins/ec2.py
```diff
@@ -49,161 +54,138 @@ def enrich_resource(event):
...
+            logger.warning(f"Unexpected ec2 resource type {event['resource_id']}")
             return
 
-    except ClientError as err:
-        if "AccessDenied" in str(err):
-            logger.error(f'Access Denied trying to enrich {resource.resource_id}. Was our access removed?')
-            data = {}
-            resource_state = reactor.constants.RESOURCE_STATE_NO_ACCESS
-            resource_status = reactor.constants.RESOURCE_STATUS_UNKNOWN
+    except botocore.exceptions.ClientError as error:
```
killthrush: After we build a few plugins, we should have a decorator do this stuff for us.

reactor/features/resources/plugins/ec2.py
```diff
@@ -49,161 +54,138 @@ def enrich_resource(event):
...
+def _process_vpc(event, aws):
+    vpc = aws.resource.Vpc(event['cloud_local_id'])
+    vpc.load()
+    resource_state = vpc.state
+    data = vpc.meta.data
+    return data, resource_state
 
 
-def configure_flow_log(ec2_provider, vpc_id):
+def configure_flow_log(ec2_provider, vpc_id, kinesis=False):
```
killthrush: This code totally does not belong here - can you move it somewhere else please?

reactor/features/resources/plugins/ec2.py
```diff
@@ -215,57 +197,73 @@ def create_vpc_flow_log(vpc_id, logs_provider, ec2_provider):
...
+    if response['Unsuccessful']:
+        logger.warning(f'Could not create VPC Flow Logs for {vpc_id} ({response})')
+        return None
+    else:
+        logger.info(f'Created VPC Flow Log: {log_group_name}')
         return log_group_name
 
 
-def subscribe_to_flow_log(vpc_id, logs_client):
+def subscribe_lambda_to_flow_log(log_group_name, logs_client):
```
killthrush: this should also move

reactor/features/resources/reconcile.py
```diff
@@ -0,0 +1,150 @@
...
+from reactor.utils import create_random_string
+from reactor.utils.serializers import serialize_dictionary
+import reactor.common.czrn as czrn
+
+logger = setup_logging('reactor')
+resources_dbh = dynamodb.CZReactorResources()
+
+
+@provisioned_throughput_retry(max_attempts=3)
+def reconcile(resource_id, timestamp, data, disposition=None, resource_state=None, cz_metadata=None):
```
killthrush: good call moving this out

reactor/features/resources/reconcile.py
```diff
@@ -0,0 +1,150 @@
...
+
+    return processed_count
+
+
+def update_resource(resource_id, timestamp, data):
+    update_expression_keys = []
+    update_expression_values = {}
+    update_expression_attribute_names = {}
+    for x, key in enumerate(data.keys(), start=1):
+        if is_dynamodb_reserved_word(key):
```
killthrush: What's the situation where you think we would have a problem with keys?

reactor/utils/czrn.py
```diff
@@ -9,6 +9,9 @@
 logger = setup_logging('reactor')
 
 
+# Everything in this file is used now only by VPC flow logs, will all be replaced when vpc flow logs are replaced with
+# kinesis
```
killthrush: thanks for adding this

serverless.yml
```diff
@@ -132,7 +165,24 @@ functions:
             Fn::GetAtt:
               - CzReactorResources
               - StreamArn
+  #enrichment
+#  resources-enrich-plugin-ec2:
+#    handler: lambda_handlers.resources__enrich_plugin_ec2
+#    memorySize: 512
+#  resources-enrich-plugin-general:
+#    handler: lambda_handlers.resources__enrich_plugin_general
+#    memorySize: 1024
```
killthrush: feel free to wipe these permanently

reactor/common/czrn.py
```diff
@@ -561,6 +565,117 @@ def create_from_arn(arn):
...
+    """
+    resource_type = 'unknown'
+    try:
+        ip = ipaddress.ip_interface(network_location)
+        resource_type = f'ipv{ip.version}-address'
+    except ValueError:
+        if network.is_valid_hostname(network_location):
+            resource_type = "host"
+        elif network.is_valid_mac_address(network_location):
+            resource_type = 'mac-address'
```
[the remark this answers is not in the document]
silvexis: not sure but i would look to refactor this whole thing in the future

reactor/common/czrn.py
```diff
@@ -561,6 +565,117 @@ def create_from_arn(arn):
...
+
+    return location, ip_metadata
+
+
+def create_from_network_location(network_location, account_id, region=None, interface_id=None):
+    """
+    Convert a input IP address or host name with or without src port and path to CZRN format
+
+    Format:
+        czrn:provider:service-type:region:account-id:resource-type:interface-id:cloud-local-id/path
```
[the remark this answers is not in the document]
silvexis: REACTOR-369

reactor/common/czrn.py
```diff
@@ -561,6 +565,117 @@ def create_from_arn(arn):
...
+    """
+    interface_id = interface_id or ""
+    provider, _ = _determine_ip_address_location(network_location)
+
+    return create_from_components(provider=provider,
+                                  region=region,
+                                  service_type="cz-network-location",
+                                  resource_type=_determine_network_location_resource_type(network_location),
+                                  account_id=account_id,
+                                  cloud_local_id=f"{interface_id}:{network_location}")
```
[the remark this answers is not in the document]
silvexis: i swapped the colon for a pipe for now, but whole function needs help in future fix

reactor/common/czrn.py
```diff
@@ -676,3 +791,27 @@ def extract_account_id(czrn):
...
+
+    Returns:
+        set:
+    """
+    # old way, do a bunch of stuff to walk the dictionary and find arn keys:
+    # return {value for key, value in extract_arns_with_details_from_event(raw_event)}
+
+    # New way...extract all ARNs with this one neat trick
+    event_text = json.dumps(raw_event)
+    m = re.findall(r'(arn:aws:[a-zA-Z0-9\-_/.:]+?)\"', event_text)
```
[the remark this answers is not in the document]
silvexis: i tried, it did't work for reasons

reactor/constants.py
```diff
@@ -34,19 +30,26 @@ def __str__(self):
...
-RESOURCE_DISPOSITION_REFRESHED = 'refreshed'            # The resource ID just expired and was refreshed
-RESOURCE_DISPOSITION_IDENTIFIED = 'identified'          # We've done a checked and confirmed the resource
-RESOURCE_DISPOSITION_PENDING = 'pending'                # We're sure this is a resource, but confirmation is needed
-RESOURCE_DISPOSITION_NOT_SUPPORTED = 'not_supported'    # We tried to identify the resource but couldn't
-RESOURCE_DISPOSITION_ERROR = 'error'                    # unknown error attempting to process this resource
+
+RESOURCE_DISPOSITION_NEW = 'new'                            # The resource ID was just discovered, has not been enriched
+RESOURCE_DISPOSITION_IDENTIFIED = 'identified'              # We've enriched the resource
+RESOURCE_DISPOSITION_STALE = 'stale'                        # Resource is stale, will be final on next ttl cycle
+RESOURCE_DISPOSITION_FINAL = 'final'                        # Resource is gone, will never return
```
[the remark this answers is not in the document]
silvexis: actually the comment is wrong, final doesn't mean the resource is gone, final means that we don't ever expect to ever hear anything about this resource ever again. It might be because it went away most of the time but that's not the only thing

reactor/constants.py
```diff
@@ -34,19 +30,26 @@ def __str__(self):
...
+RESOURCE_STATUS_FINAL = 'final'
+RESOURCE_STATUS_NO_ACCESS = 'no-access'
+RESOURCE_STATUS_CROSS_ACCOUNT = 'cross-account'
+
 
 RESOURCE_STATE_UNKNOWN = 'unknown'
-RESOURCE_STATE_FINAL = 'final'
-RESOURCE_STATE_NO_ACCESS = 'no-access'
+RESOURCE_STATE_GONE = 'gone'
+RESOURCE_STATE_PRESENT = 'present'
```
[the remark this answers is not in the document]
silvexis: yeah it's gone, these constants need to go, i'll zap em

reactor/database/models/dynamodb/cz_reactor_resources.py
```diff
@@ -61,20 +65,19 @@ def write(self, records):
     def create_new_resources(self, resource_ids, source_environment_id, cloud_account_id):
         resource_records = []
         for resource_id, timestamp in resource_ids.items():
-            new_resource = create_resource_record(service_type=get_service_type_from_arn(resource_id),
-                                                  resource_name=get_object_name_from_arn(resource_id),
-                                                  resource_type=get_resource_type_from_arn(resource_id),
-                                                  region=get_region_from_arn(resource_id),
+            new_resource = create_resource_record(service_type=czrn.extract_service_type(resource_id),
```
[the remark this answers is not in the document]
silvexis: probably ;)

PR #68, opened 2018-03-27
REACTOR-362 - Extend CZ-NEF to store Access Key Id Used and Access Key Id Granted

tiyberius:
### Description

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


tiyberius:
Ah, @imcgaunn I'm glad you asked about `id_granted` vs `id_used`. That is a very important distinction, so I'd like to find the correct terminology. The Access Key Id `granted` is the Access Key Id given to a user or an account by AWS when it authenticates successfully. And when that user or account USES that Access Key Id later, that is the `id_used`

reactor/aws/tools.py
```diff
@@ -589,3 +592,59 @@ def get_reactor_functions(namespace):
...
+
+
+def get_access_key_id_granted_in_event(event):
+    """
+    >>> get_access_key_id_granted_in_event({'responseElements': {'credentials': {'accessKeyId': 'flargy'}}})
+    'flargy'
+    >>> get_access_key_id_granted_in_event(None)
+    Traceback (most recent call last):
+    ...
+    TypeError: 'NoneType' object is not subscriptable
```
killthrush: this can also raise key errors.  Add a test case?

reactor/aws/tools.py
```diff
@@ -589,3 +592,59 @@ def get_reactor_functions(namespace):
...
+    access_key_id = event['responseElements']['credentials']['accessKeyId']
+    return access_key_id
+
+
+def get_access_key_id_granted_in_event_if_exists(event):
+    """
+    >>> get_access_key_id_granted_in_event_if_exists({'responseElements': {'credentials': {'accessKeyId': 'flargy'}}})
+    'flargy'
+    >>> get_access_key_id_granted_in_event_if_exists({})
+    >>> get_access_key_id_granted_in_event_if_exists(None)
```
killthrush: So if there's no value on the line below the test case/example that means 'this function returns None'?

reactor/features/simplequery/adapters.py
```diff
@@ -53,6 +53,8 @@
     externalId: STRING >,
   event_resources ARRAY <
     STRING >,
+  access_key_id_used STRING,
+  access_key_id_granted STRING,
```
killthrush: Do we need to add these to CZ-NEF and to aurora?

reactor/aws/tools.py
tiyberius: That is correct. A bit confusing, but that's what happens in the Python interpreter, so it's what happens here too

reactor/features/simplequery/adapters.py
tiyberius: Yes! We totally need to. What do you think about the column names ... do they make sense?

reactor/aws/tools.py
```diff
@@ -589,3 +592,59 @@ def get_reactor_functions(namespace):
...
+    >>> get_access_key_id_used_in_event_if_exists({'userIdentity': {'accessKeyId': 'flargy'}})
+    'flargy'
+    >>> get_access_key_id_used_in_event_if_exists({})
+    >>> get_access_key_id_used_in_event_if_exists(None)
+    """
+    access_key_id = None
+    try:
+        access_key_id = get_access_key_id_used_in_event(event)
+    except (KeyError, TypeError):
+        logger.debug(f'Unable to get {access_key_id} used in event')
```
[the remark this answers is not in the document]
tiyberius: done

reactor/aws/tools.py
tiyberius: great catch, will be fixed in next update

PR #72, opened 2018-04-09
Feature/reactor 354 - solidify resource type extraction from ARNs

killthrush:
### Description

Naive approaches to resource type extraction from ARNs kind of fall apart when considering the space of all possible ARNs (see https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html).  Beef this up with comprehensive unit tests.  Also normalize the extracted resource type; in some cases the format of the resource type isn't that great as-is.  Abbreviations and inconsistent formatting is common and sometimes the type name isn't event present.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2016-present, CloudZero, Inc. All rights reserved.
+# Licensed under the BSD-style license. See LICENSE file in the project root for full license information.
+
+"""
+Contains useful functions for processing ARNs of all shapes and sizes.
+"""
```
killthrush: new module; this anticipated some of Ian's feedback in reactor-355.  ARN processing is complex enough to warrant  its own module.

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+
+import re
+import lambda_tools as lt
+
+
+_ARN_MATCH_REGEX = re.compile('^arn:aws:([a-z0-9\-]+):([a-z0-9\-]*):(\d{12})*:(.+$)$')
```
killthrush: All ARNs follow this format.  The resource section (i.e. the 5th capturing group) is where all the complexity lies

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+
+logger = lt.setup_logging('reactor')
+
+
+# We can't use boto to reliably give us a list of all the service type strings without extra work.
+# The set of strings that can appear in arns is different than what can be used with the boto API,
+# however there is significant overlap.
+# TODO: establish mappings for these 2 domains
```
killthrush: `_service_types_in_arns` below is necessary for extracting resource types.  At present `_service_types_unique_to_boto` and `_service_types_unique_to_arns` are only for reference.

reactor/aws/tools.py
```diff
@@ -86,10 +79,10 @@ def get_object_name_from_arn(arn):
     """
     Get the object name from an ARN
 
-    >>> get_object_name_from_arn('arn:aws:ec2:us-east-1:123456789012:instance/i-0ebaff919fe99b7a5')
+    >>> get_object_name_from_arn'ec2:us-east-1:123456789012:instance/i-0ebaff919fe99b7a5')
```
killthrush: hmm, these changes were not intentional.  I'll back them out

test/unit/aws/test_unit_tools.py
```diff
@@ -2,35 +2,52 @@
...
+    assert tools.get_resource_type_from_arn(None) is None
+    assert tools.get_resource_type_from_arn('') is None
+    assert tools.get_resource_type_from_arn('uhm::') is None
+    assert tools.get_resource_type_from_arn('arn:microsoft:ohno:::') is None
+    assert tools.get_resource_type_from_arn('arn:aws:states:') is None
+    assert tools.get_resource_type_from_arn('::kinesis:::') is None
+    assert tools.get_resource_type_from_arn('arn:aws:barf:us-east-42:123456789012:something/something-id') is None
+
+
+def test_get_resource_type_from_arn_valid_input_maybe_kinda_question_mark():
```
killthrush: This test is small in stature but a giant in usefulness

test/data/all_arns.py
```diff
@@ -0,0 +1,231 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2016-present, CloudZero, Inc. All rights reserved.
+# Licensed under the BSD-style license. See LICENSE file in the project root for full license information.
+
+"""
+Test file containing a massive list of sample resource ARNs from all supported AWS services.
+Each ARN is packaged with important elements that have been extracted from the ARN like service and resource type.
+See https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
+"""
```
killthrush: This test data file contains about 220 different resource type ARNs derived from AWS's documentation.

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+}
+
+_service_types_unique_to_arns = {
+    'trustedadvisor', 'elasticfilesystem', 'a4b', 'states', 'mobilehub', 'mediatailor', 'execute-api',
+    'artifact', 'mobiletargeting', 'elasticloadbalancing'
+}
+
+
+# As part of extracting resource types, we also transform/map certain strings to more readable
+# normalized equivalents.  ARNs vary in quality for self-documentation.
```
killthrush: At the end of resource type extraction, we'll have a nice clean expressive resource type identifier.

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+    'statemachine': 'state-machine',
+    'subgrp': 'subnet-group',
+    'subnetgroup': 'subnet-group',
+    'threatintelset': 'threat-intel-set',
+    'vaults': 'vault',
+    'webacl': 'web-acl',
+}
+
+
+def get_resource_type_from_arn(arn):
```
killthrush: This is the function that does all the heavy lifting.  It's a mixture of custom function calls and regular parsing logic for simple cases.

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+        service, _, _, resource = arn_match.groups()
+        if service not in _service_types_in_arns:
+            raise ValueError(f'Unknown service {service}')
+        custom_handlers_map = _get_resource_type_custom_handlers()
+        if service in custom_handlers_map:
+            return custom_handlers_map[service](resource)
+        elif '/' in resource:
+            return _normalize_resource_type_name(resource.split('/')[0])
+        elif ':' in resource:
+            return _normalize_resource_type_name(resource.split(':')[0])
```
killthrush: ARN resource components are delimited by slashes or colons.  In some cases, colons and slashes appear together - but this is covered by custom extractor functions.

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+    if not arn:
+        return None
+    try:
+        arn_match = re.fullmatch(_ARN_MATCH_REGEX, arn)
+        if not arn_match or len(arn_match.groups()) != 4:
+            raise ValueError('ARN did not match standard regex')
+
+        service, _, _, resource = arn_match.groups()
+        if service not in _service_types_in_arns:
+            raise ValueError(f'Unknown service {service}')
```
silvexis: Include arn in raise?

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+        custom_handlers_map = _get_resource_type_custom_handlers()
+        if service in custom_handlers_map:
+            return custom_handlers_map[service](resource)
+        elif '/' in resource:
+            return _normalize_resource_type_name(resource.split('/')[0])
+        elif ':' in resource:
+            return _normalize_resource_type_name(resource.split(':')[0])
+        else:
+            return None
```
silvexis: should this raise a value error instead of returning None? If we can't extract a resource type thats a but or at the very lest something we need to know

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+    except Exception as e:
+        logger.exception(e)
+        logger.warning(f'Invalid ARN: {arn}')
+        return None
```
silvexis: raise ValueError?

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+        return 'query-field'
+    if '/types/Mutation/' in arn_resource:
+        return 'mutation-field'
+    if '/types/Subscription/' in arn_resource:
+        return 'subscription-field'
+    return None
+
+
+def _extract_logs_resource_types(arn_resource):
+    if arn_resource.startswith('log-group') and 'log-stream' not in arn_resource:
```
silvexis: short circuit with startswith at top and simplify if's? 
```
if not arn_resource.startswith('log-group'):
     return None
```

reactor/aws/tools.py
```diff
@@ -158,40 +151,22 @@ def get_service_type_from_arn(arn):
...
+    # TODO: this wrapper is here to make the merge easier - replace call sites with this version
+    return arn_utils.get_resource_type_from_arn(arn)
 
 
 def get_account_id_from_arn(arn):
     """
     Get Account ID from ARN
 
-    >>> get_account_id_from_arn('arn:aws:ec2:us-east-1:526216611803:instance/i-0ebaff919fe99b7a5')
+    >>> get_account_id_from_arn'ec2:us-east-1:526216611803:instance/i-0ebaff919fe99b7a5')
```
silvexis: fix

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+    if arn_resource.startswith('listener-rule/net'):
+        return 'net-load-balancer-listener-rule'
+    if arn_resource.startswith('targetgroup'):
+        return 'target-group'
+    return None
+
+
+def _extract_iam_resource_types(arn_resource):
+    if arn_resource == 'root':
+        return 'root'
```
silvexis: I think this should return 'aws-account' for root

test/data/all_arns.py
silvexis: You are missing one, `'arn:aws:iam::account-id:role/service-role/config-role-us-east-1'` which should return 'service-role' as the type. This ARN does not appear on the AWS web site, but has been observed in prod

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+            return custom_handlers_map[service](resource)
+        elif '/' in resource:
+            return _normalize_resource_type_name(resource.split('/')[0])
+        elif ':' in resource:
+            return _normalize_resource_type_name(resource.split(':')[0])
+        else:
+            return None
+    except Exception as e:
+        logger.exception(e)
+        logger.warning(f'Invalid ARN: {arn}')
```
silvexis: logger.error

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,301 @@
...
+    Get a normalized resource type from an ARN
+
+    Args:
+        arn: (str) - the arn to deconstruct
+
+    Returns:
+        str - a canonical resource type name
+    """
+    if not arn:
+        return None
```
silvexis: raise

reactor/aws/arn_utils.py
killthrush: I could go either way with this - I think the decision is going to depend on whether or not the caller is equipped to deal with a raised exception.  If you think the answer is "yes" I'll go ahead and change this

reactor/aws/arn_utils.py
killthrush: Is this a fatal error condition?  If it is, I'll change this

reactor/aws/arn_utils.py
killthrush: I'll try simplifying this - thankfully I have tests =)

reactor/aws/arn_utils.py
killthrush: per our discussion, we'll take a look at general error handling in a later ticket

reactor/aws/arn_utils.py
killthrush: This worked, and applied to some of these other functions

test/data/all_arns.py
silvexis: This service-role still needs to be added

reactor/aws/arn_utils.py
silvexis: anytime i see logger.exception i assume it's an error, logger.exception logs as error, makes no sense to log as a warning right after

reactor/aws/arn_utils.py
silvexis: yeah per discussion, it's fine as is

reactor/aws/arn_utils.py
killthrush: I think that's more of a confusing feature of the logging library more than anything else.  I'm interested in the stack trace here and I think that's still valid.  I'll change this to log just the stack trace so it's less confusing

test/data/all_arns.py
killthrush: Oh, sorry I didn't see this because it was in a file that was collapsed.  will fix

reactor/aws/arn_utils.py
```diff
@@ -0,0 +1,308 @@
...
+        if service in custom_handlers_map:
+            return custom_handlers_map[service](resource)
+        elif '/' in resource:
+            return _normalize_resource_type_name(resource.split('/')[0])
+        elif ':' in resource:
+            return _normalize_resource_type_name(resource.split(':')[0])
+        else:
+            return None
+    except Exception as e:
+        logger.warning(f'Invalid ARN: {arn}', exc_info=True)
```
killthrush: AHA!  this way is better ^^^.  In fact you can use this technique anywhere and can get rid of the extra `exception` calls.  UNder the hood, it's just using the built-in python facilities for stack traces (`sys.exc_info()`)

PR #73, opened 2018-04-19
Feature/reactor 361 - czrn creation and component functions

killthrush:
### Description

Provide a first pass at an implementation for CZRNs.  Includes functions to create them from components (like account ID, resource type, etc), create them from ARNs, and also to extract individual components from them.

Also, this required implementation of a function to reliably extract a "cloud local ID" from ARNs; this information will be used during resource enrichment. 

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


Makefile
```diff
@@ -70,7 +70,7 @@ lint:                                    ## lints the code via adherence to PEP8
 
 
 lint-fix:								                 ## fixes the code in place so that it will pass `make lint`
-	autopep8 --ignore E265,E266,E402 --in-place --recursive --max-line-length=120 --exclude vendored .
+	autopep8 --ignore E265,E266,E402 --in-place --recursive --max-line-length=400 --exclude vendored .
```
killthrush: Thought we still have the linter set up to use 120, this prevents the autoformatted from wrapping lines unless they are really long.  You'll see why in a bit.

reactor/aws/arn_utils.py
```diff
@@ -1,316 +0,0 @@
-# -*- coding: utf-8 -*-
-# Copyright (c) 2016-present, CloudZero, Inc. All rights reserved.
-# Licensed under the BSD-style license. See LICENSE file in the project root for full license information.
-
-"""
-Contains useful functions for processing ARNs of all shapes and sizes.
-"""
```
killthrush: moved this whole file into `common/czrn`.  I'm slowly refactoring to the domain-centric model employed by core-web.  This preexisting code for resource type extraction did not change.

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+"""
+
+import ipaddress
+import re
+import lambda_tools as lt
+
+from reactor.utils.network import is_valid_mac_address, is_valid_hostname
+
+_ARN_MATCH_REGEX = re.compile('^arn:aws:([a-z0-9\-]+):([a-z0-9\-]*):(\d{12})*:(.+$)$')
+_CZRN_MATCH_REGEX = re.compile('^czrn:aws:([a-z0-9\-]+):([a-z0-9\-]*):(\d{12})*:([a-z0-9\-]+):(.+$)$')
```
killthrush: new regex for CZRNs

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+        'serverlessrepo': _extract_serverlessrepo_resource_types,
+        'sns': _extract_sns_resource_types,
+        'sqs': lambda r: 'queue',
+        's3': _extract_s3_resource_types,
+        'swf': lambda r: 'domain',
+        'storagegateway': _extract_storage_gateway_resource_types,
+    }
+
+
+def get_cloud_local_id_from_arn(arn):
```
killthrush: new code starts here and continues till the end of the file.  I'm probably going to split this module up into pieces, but how I do that is still open to discussion.  I like the idea of having service-specific modules so I can keep custom functions together without overwhelming the reader.

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+    try:
+        service, _, _, resource = _get_arn_components(arn)
+        resource_type = get_resource_type_from_arn(arn)
+        custom_handlers_map = _get_cloud_local_id_custom_handlers()
+        if service in custom_handlers_map:
+            return custom_handlers_map[service](resource, resource_type)
+        elif '/' in resource:
+            return resource.split('/')[1]
+        elif ':' in resource:
+            return resource.split(':')[1]
```
killthrush: like with resource types, a lot of cloud-local-ids can be extracted from an ARN without a fuss. But some custom code is still needed.  Additionally, having a resource type on hand makes our lives easier, so I made it an argument to the extraction functions.

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+        'elasticbeanstalk': _extract_elasticbeanstalk_cloud_local_ids,
+        'elasticloadbalancing': lambda r, t: r.split('/')[-1:][0],
+        'iam': _extract_iam_cloud_local_ids,
+        'sts': _extract_sts_cloud_local_ids,
+        'kinesisvideo': lambda r, t: '/'.join(r.split('/')[1:]),
+        'lambda': lambda r, t: '/'.join(r.split(':')[1:]),
+        'mediastore': lambda r, t: '/'.join(r.split('/')[1:]),
+        'organizations': _extract_organizations_cloud_local_ids,
+        'mq': lambda r, t: r.split(':')[-1:][0],
+        'mobiletargeting': lambda r, t: r.split('/')[-1:][0],
```
killthrush: got some good slicing practice on this one!

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+def _extract_storagegateway_cloud_local_ids(arn_resource, resource_type):
+    parts = arn_resource.split('/')
+    if resource_type in {'gateway', 'share', 'tape'}:
+        return parts[1]
+    elif resource_type in {'scsi-target', 'volume', 'device'}:
+        return parts[1] + '/' + parts[3]
+    return None
+
+
+def create_from_components(service_type, resource_type, cloud_local_id, provider=None, region=None, account_id=None):
```
killthrush: here's where the fun starts - making a new CZRN

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+    """
+    if not all({service_type, resource_type, cloud_local_id}):
+        return None
+    provider = provider or 'aws'
+    region = region or ''
+    account_id = account_id or ''
+    return f'czrn:{provider}:{service_type}:{region}:{account_id}:{resource_type}:{cloud_local_id}'
+
+
+def create_from_arn(arn):
```
killthrush: ...and a shortcut to create from an ARN if we have one

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+    if not arn:
+        return None
+
+    service, region, account_id, _ = _get_arn_components(arn)
+    resource_type = get_resource_type_from_arn(arn)
+    cloud_local_id = get_cloud_local_id_from_arn(arn)
+    return create_from_components(service, resource_type, cloud_local_id, region=region, account_id=account_id)
+
+
+def extract_all_components(czrn):
```
killthrush: and functions for the inverse

test/data/all_arns.py
```diff
@@ -9,223 +9,304 @@
...
-    ('arn:aws:waf::123456789012:webacl/3bffd3ed-fa2e-445e-869f-a6a7cf153fd3', 'waf', 'web-acl', ),
-    ('arn:aws:waf::123456789012:ipset/3f74bd8c-f046-4970-a1a7-41aa52e05480', 'waf', 'ip-set', ),
-    ('arn:aws:waf::123456789012:bytematchset/d131bc0b-57be-4536-af1d-4894fd28acc4', 'waf', 'byte-match-set', ),
-    ('arn:aws:waf-regional:us-east-1:123456789012:rule/41b5b052-1e4a-426b-8149-3595be6342c2', 'waf-regional', 'rule', ),
-    ('arn:aws:waf-regional:us-east-1:123456789012:webacl/3bffd3ed-fa2e-445e-869f-a6a7cf153fd3', 'waf-regional', 'web-acl', ),
-    ('arn:aws:waf-regional:us-east-1:123456789012:ipset/3f74bd8c-f046-4970-a1a7-41aa52e05480', 'waf-regional', 'ip-set', ),
-    ('arn:aws:waf-regional:us-east-1:123456789012:bytematchset/d131bc0b-57be-4536-af1d-4894fd28acc4', 'waf-regional', 'byte-match-set')
+    # a4b examples are incomplete
+    ('arn:aws:a4b:us-east-1:123456789012:room/7315ffdf0eeb874dc4ab8a546e8b70ec/5f90e5d608b6baa9c88db56654aef158', 'a4b', 'room', '7315ffdf0eeb874dc4ab8a546e8b70ec/5f90e5d608b6baa9c88db56654aef158', ),
+
```
killthrush: OK, so this file didn't really change except that I added new test values for cloud-local-id.  Some of the values are probably wrong and we'll find out later =)

This file was why I turned off the long-line wrapping - this file is impossible to read with wraps

test/unit/common/test_czrn.py
```diff
@@ -0,0 +1,144 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2016-present, CloudZero, Inc. All rights reserved.
+# Licensed under the BSD-style license. See LICENSE file in the project root for full license information.
+
+import pytest
+import test.data.all_arns as arns
+import reactor.common.czrn as czrn
+
+
+def test_get_resource_type_from_arn_invalid_input():
```
killthrush: moved some tests here from the existing module...

test/unit/common/test_czrn.py
```diff
@@ -0,0 +1,144 @@
...
+
+def test_get_resource_type_from_arn_valid_input():
+    # NOTE: The sample ARNs here were derived from AWS's documentation, however there were a few cases where I think
+    # the published examples were wrong.  So this is about as reliable as it could be, but might still be inaccurate.
+    for arn, _, expected_resource_type, _, in arns.ALL_ARNS_WITH_DETAILS:
+        actual_resource_type = czrn.get_resource_type_from_arn(arn)
+        assert actual_resource_type == expected_resource_type
+
+
+def test_get_cloud_local_id_from_arn_invalid_input():
```
killthrush: ...and added a bunch of new ones

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+    lowercase_resource = arn_resource.lower()
+    m = re.fullmatch('^([a-z\-_0-9/]+).*', lowercase_resource)
+    name = m.groups()[0]
+    if name in _transformable_resource_type_names:
+        return _transformable_resource_type_names[name]
+    return name
+
+
+def _get_resource_type_custom_handlers():
+    """
```
imcgaunn: does this need to be a function?

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+        return parts[1]
+    elif resource_type in {'scsi-target', 'volume', 'device'}:
+        return parts[1] + '/' + parts[3]
+    return None
+
+
+def create_from_components(service_type, resource_type, cloud_local_id, provider=None, region=None, account_id=None):
+    """
+    Create a CZRN identifier given its components, some of which are optional.
+    All resources must have a CZRN that complies with this format.
```
silvexis: add a few example CZRN's in the docstring to give the reader an idea of what they will get? maybe as doctests?

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+        'sqs': lambda r: 'queue',
+        's3': _extract_s3_resource_types,
+        'swf': lambda r: 'domain',
+        'storagegateway': _extract_storage_gateway_resource_types,
+    }
+
+
+def get_cloud_local_id_from_arn(arn):
+    """
+    Get the resource-specific cloud-local-id from an ARN.
```
imcgaunn: excellent documentation throughout this work.

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+        else:
+            return None
+    except Exception as e:
+        logger.warning(f'Invalid ARN: {arn}', exc_info=True)
+        return None
+
+
+def _get_cloud_local_id_custom_handlers():
+    """
+    Get a mapping of service types to custom functions used for extracting cloud-local-IDs from an ARN.
```
imcgaunn: this function is hard to understand due to the slicing / indices. It's possible this could be made easier with names for the indices used in the slices?

imcgaunn: or names for the slices themselves?

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+    if not czrn_match or len(czrn_match.groups()) != 5:
+        raise ValueError(f'CZRN {czrn} did not match standard regex')
+    service, region, account_id, resource_type, cloud_local_id = czrn_match.groups()
+    if service not in _service_types_in_arns:
+        raise ValueError(f'Unknown service {service} in arn: {arn}')
+    return service, region or None, account_id or None, resource_type, cloud_local_id
+
+
+def extract_service_type(czrn):
+    service, _, _, _, _ = extract_all_components(czrn)
```
imcgaunn: I always liked this syntax.

Makefile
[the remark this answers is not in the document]
killthrush: autopep8 is a formatting tool that in some respects is configured similarly.  I usually run it from make during development.  Re: formatting exceptions, I'm not sure how to do it with this tool.  But I will look a bit harder.  I may have to exclude files here in the makefile but that might not be so bad

reactor/common/czrn.py
killthrush: I had doctests originally but couldn't get them to run as part of a test suite for some reason.  I can put them back as examples though, that would be no trouble at all.

Makefile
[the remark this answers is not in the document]
killthrush: 400 is arbitrary, probably don't need it, you're right.  However I'm going to try first to just ignore that one file instead.  Usually the autowrapping is a good thing

reactor/common/czrn.py
killthrush: Yeah I forgot you could do this in python.  I first learned to do this in elixir

reactor/common/czrn.py
killthrush: Yeah, I think I'm going to ditch the anonymous functions eventually in favor of functions with decorators and some module magic.  The tests will make the change quite safe.  I think I'm going to do this only after we've gotten a few enrichments off the ground

reactor/common/czrn.py
killthrush: Technically no, but I made it one because I didn't want to create this big dict at the top of the file.  which... I did... anyway... hmmm well, at least as a function it can have a nice docstring!  Per your other point though, I think this eventually is going to be refactored away and be compiled dynamically.  As we do more and more of this sort of thing, it will be nice to just drop in a module, give it some decorators and BAM! you now have support for a new service

reactor/common/czrn.py
```diff
@@ -0,0 +1,615 @@
...
+    'trustedadvisor', 'elasticfilesystem', 'a4b', 'states', 'mobilehub', 'mediatailor', 'execute-api',
+    'artifact', 'mobiletargeting', 'elasticloadbalancing'
+}
+
+
+# As part of extracting resource types, we also transform/map certain strings to more readable
+# normalized equivalents.  ARNs vary in quality for self-documentation.
+_transformable_resource_type_names = {
+    'applicationversion': 'application-version',
+    'autoscalinggroup': 'group',
```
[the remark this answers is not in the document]
killthrush: Resource types are always namespaced in CZRN - you can't have one in isolation

Makefile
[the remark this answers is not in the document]
killthrush: 2 different tools, 2 different use cases.  Flake8 is a linter, autopep8 is a refarmatting tool.  They actually use the same lib under the hood (`pycodestyle`, formerly `pep8`).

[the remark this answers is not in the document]
killthrush: we never want to run a reformatting tool during a build, that's just  plain nuts =)

reactor/common/czrn.py
killthrush: still wasn't able to get it to work - can you show me what you're doing?

Makefile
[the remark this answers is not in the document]
killthrush: Decided to keep this in.  The long ARN strings really push the limits of good taste in python, and pycodestyle (the lib behind autopep8 and flake8) is embroiled in a religious war about per-file overrides to actually fix this.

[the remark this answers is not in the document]
killthrush: Unfortunately i don't have a lot of good options for this sort of thing.  It's Amazon's fault for requiring such long strings!

PR #77, opened 2018-04-30
HOTFIX: Disable resource count

silvexis:
### Description
Fix issue with counts and DynamoDB stream performance issues


reactor/features/ingest.py
```diff
@@ -337,9 +337,8 @@ def normalize_cloudtrail_events(events, environments, event_stats, message_sourc
 
                     total_res_processed += 1
 
-            else:
-                logger.warning(f'Invalid resource received: {resource}\n'
-                               f'Event: {cloudtrail_event.normalized_event}')
+            except ValueError:
```
killthrush: would do a catch-all here until we can refactor this.  Also please add a comment

reactor/features/resources/curate.py
```diff
@@ -97,7 +98,12 @@ def curate(records, reactor_account_id, namespace):
         raw_image = new_image or old_image
         if not raw_image:
             raise ValueError('Invalid dynamodb stream record: {}'.format(record))
-        data_image = lambda_utils.deserialize_dynamo_data(raw_image)
+
+        try:
+            data_image = lambda_utils.deserialize_dynamo_data(raw_image)
+        except Exception as error:
+            logger.error(f'Failed to deserialize DynamoDB data {error}, skipping record')
```
killthrush: Add debug log for the batch contents here? otherwise we'll probably never be able to fix them.

reactor/features/resources/reconcile.py
```diff
@@ -92,7 +97,7 @@ def reconcile(resource):
     processed_count = 0
 
     for version in previous_versions['Items']:
-        version_count += version.get('count', 1)
+        # version_count += version.get('count', 1)
```
killthrush: Please add a comment above this line explaining why it's commented out

reactor/features/resources/get.py
```diff
@@ -46,11 +46,11 @@ def get_resources(cloud_account_id):
                                    "timestamp": item['timestamp']})
     logger.info(f"Purged {purge_count} of {len(stale.get('Items'))} stale resources")
 
-    # Purge old ARN res_id's
+    # Purge pending res_id's
     done = False
     kwargs = {}
     while not done:
-        stale = res_dbh.table.scan(FilterExpression=Attr('res_id').begins_with('arn'),
+        stale = res_dbh.table.scan(FilterExpression=Attr('disposition').eq('pending'),
```
killthrush: +100 internet points for recycling =)

reactor/features/ingest.py
silvexis: it's pretty reliably a ValueError when it goes wrong, which I think is going to always happen because we are parsing potentially garbage from AWS and we know now that ARN's are a very lose "standard", bu I can catch both just to be safe

PR #79, opened 2018-05-04
Feature/reactor 288 - enhance instance plugin and establish test patterns

killthrush:
### Description

As a prerequisite for writing a bunch of plugins that support VPC Flow Log event sources, improve the quality of the `ec2.instance` enrichment plugin and add unit tests.  This should set a solid precedent for future work.

Also improve the documentation and clean up some cruft that was hanging around from the original CZRNv1 feature.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,172 +1,228 @@
...
+    instance_data = aws.client.describe_instances(MaxResults=1000)['Reservations'][0]['Instances']  # TODO: fix this
+    logger.debug(instance_data)
+    return instance_data
+
+
+def _get_instance_properties(instance_data):
+    interesting_keys = {
+        'AmiLaunchIndex', 'Architecture', 'EbsOptimized', 'InstanceId', 'InstanceType',
+        'KernelId', 'KeyName', 'LaunchTime', 'VpcId', 'Platform', 'RamdiskId', 'SourceDestCheck',
+        'VirtualizationType', 'EnaSupport', 'Hypervisor',
```
killthrush: Before we were just dumping the boto call to properties without any real thinking.  These are the ones that really matter.  Some of the others are actually explicit relationships that we need to process as CZRNs, and others are "rolled-up" derived relationships generated by amazon.

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,172 +1,228 @@
...
+        'monitoring_state': instance_data['Monitoring']['State'],
+        'state_code': instance_data['State']['Code'],
+        'state_name': instance_data['State']['Name'],
+        'state_transition_reason': instance_data['StateTransitionReason'],
+
+        # TODO: figure out whether we should treat VPC and AZ as first-class citizens that all resources get
+        # and that are defined in CZ-NRF (i.e. the "basis").  These may be handled better as derived relationships.
+        # Also not sure about what to do for AZ if there's no placement group.
+        'availability_zone': instance_data['Placement']['AvailabilityZone'],
+        'vpc_id': instance_data['VpcId'],
```
killthrush: For now, we'll still treat these two as valid properties.  Technically, these are derived relationships that should be resolved in Core.

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,172 +1,228 @@
...
+
+    subnet_id = instance_data.get('SubnetId')
+    if subnet_id:
+        subnet_czrn = czrn.create_from_components(service_type='ec2', resource_type='subnet',
+                                                  cloud_local_id=subnet_id,
+                                                  region=region, account_id=account_id)
+        relationships.append(subnet_czrn)
+
+    logger.debug(f'ec2.instance: relationships found: {relationships}')
+    return relationships
```
killthrush: This is the set of important first-degree relationships that we really should capture.  We're still doing subnet and VPC here, mainly for pragmatic reasons - we need these resources for flow logs.

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -228,157 +287,3 @@ def _process_vpc(event, aws):
     resource_state = vpc.state
     data = vpc.meta.data
     return data, resource_state
-
-
-def configure_flow_log(ec2_provider, vpc_id, kinesis=False):
```
killthrush: moved to its own module

reactor/features/resources/plugins/ec2/vpc.py
```diff
@@ -0,0 +1,16 @@
...
+This resource type is explicitly supported by Amazon.
+"""
+
+
+def enrich_cooldown():
+    return 30
+
+
+def enrich_resource(resource_stream_event):
+    raise NotImplementedError('AAAAAAA!')
```
killthrush: This is a stub for now - the current ticket I'm working on calls for this to be implemented, but I wanted eyes on this now

reactor/features/resources/plugins/plugin_loader.py
```diff
@@ -0,0 +1,43 @@
...
+    except ModuleNotFoundError:
+        logger.debug(f'No plugin found for {service_type}.{resource_type}, attempting to load default for service')
+
+    try:
+        module = importlib.import_module(f'reactor.features.resources.plugins.{service_type}.default')
+        logger.info(f'Loaded default plugin for service type {service_type}')
+        return module
+    except ModuleNotFoundError:
+        logger.debug(f'No default plugin found for {service_type}')
+        return None
```
killthrush: This code was from the `enrich` module originally.  Didn't really change, but we now have tests.

test/unit/features/plugins/ec2/test_instance.py
```diff
@@ -0,0 +1,291 @@
...
+import unittest.mock as mock
+import decimal as dl
+import reactor.features.resources.plugins.ec2.instance as plugin
+
+
+@pytest.fixture(scope='function')
+def context(mocker):
+    context = attrdict.AttrMap()
+    prefix = plugin.__name__
+    mock_random = mocker.patch(f'{prefix}.resource_tools.random', autospec=True)
```
killthrush: The import style we've chosen lets us do tricks like this easily =)
This is actually a second-degree import

test/unit/features/plugins/ec2/test_instance.py
```diff
@@ -0,0 +1,291 @@
...
+    mock_provider.Provider().client = context.mock_client
+    yield context
+    mocker.stopall()
+
+
+@pytest.mark.enrichment_plugin
+@pytest.mark.unit
+def test_enrich_happy_path(context):
+    response = _build_mock_response()
+    context.mock_client.describe_instances.return_value = response
```
killthrush: Our first legit plugin test!  Pure python is awesome.

test/unit/features/plugins/test_plugin_loader.py
```diff
@@ -0,0 +1,61 @@
...
+    context = attrdict.AttrMap()
+
+    context.plugin_paths = glob.glob(path.dirname(loader.__file__) + "/**/*.py")
+    context.resource_type_list = [
+        (path.basename(f)[:-3], path.dirname(f).split('/')[-1])
+        for f in context.plugin_paths
+        if path.isfile(f) and not f.endswith('__init__.py')
+    ]
+
+    yield context
```
killthrush: The tests in here verify that every plugin follows the spec and that the loader works.

reactor/database/models/dynamodb/cz_reactor_resources.py
```diff
@@ -5,24 +5,22 @@
 """
 Interface for dealing with the Reactor Resources dynamodb table, part of Resource Discovery.
 """
-import random
-from decimal import Decimal
+import decimal as dl
```
silvexis: this decreases readability from what one would expect when using the Decimal object

reactor/features/resources/plugins/ec2/default.py
```diff
@@ -0,0 +1,22 @@
...
+TODO: consider whether this should exist as a plugin at all - there might not be a real use case for this.
+"""
+
+import lambda_tools as tools
+
+logger = tools.setup_logging('reactor')
+
+
+def enrich_cooldown():
+    return 1
```
silvexis: default should be zero

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,384 +1,205 @@
...
         all_instances = _get_all_instances(ec2_provider)
 
+        # TODO: refactor this so we can load and cache lookups
         for instance in all_instances:
             data['properties'] = _get_instance_properties(instance)
-            data['relationships'] = _get_instance_relationships(instance, event['region'],
-                                                                event['cloud_account_id'])
+            data['relationships'] = _get_instance_relationships(instance, resource_stream_event['region'],
+                                                                resource_stream_event['cloud_account_id'])
             data['tags'] = _get_instance_tags(instance)
```
silvexis: i'd just move this to resource_tags right from the start, it's a new thing that we even captured this info anyway

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,384 +1,205 @@
...
-            enriched_resource['resource_state'] = instance.state['Name']
+            enriched_resource['resource_state'] = data['properties']['state_name']
+            enriched_resource['cloud_local_id'] = instance_id
+            enriched_resource['cloud_global_id'] = None  # Instances don't have ARNs
+
+            # For future use in CZ-NRF
+            enriched_resource['resource_properties'] = None
+            enriched_resource['resource_relationships'] = None
+            enriched_resource['resource_tags'] = None
+            enriched_resource['resource_present'] = None
```
silvexis: if you have all this data, you can set preset to true

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,384 +1,205 @@
...
-    ec2_provider = Provider(role_arn=aws_account_details['role_arn'],
-                            external_id=aws_account_details['external_id'],
-                            region=event['region'],
-                            service='ec2')
+    ec2_provider = provider.Provider(role_arn=aws_account_details['role_arn'],
+                                     external_id=aws_account_details['external_id'],
+                                     region=resource_stream_event['region'],
+                                     service='ec2')
     try:
         all_instances = _get_all_instances(ec2_provider)
```
silvexis: the except part of this try should come right after this get_all_instances call, or move into the private function itself.

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,384 +1,205 @@
...
 
+        # TODO: refactor this so we can load and cache lookups
```
silvexis: the lookup happens above in _get_all_instances but I don't think you want to cache that.

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,384 +1,205 @@
...
-
-
-def subscribe_lambda_to_flow_log(log_group_name, logs_client):
-    """
-    THIS FUNCTION DOES NOT WORK ACROSS ACCOUNTS :-(
+def _get_all_instances(aws):
+    # TODO: use NextToken and make this a generator.
 
-    Only Kinesis is supported for cross account cloudwatch log delivery
+    instance_data = aws.client.describe_instances(MaxResults=1000)['Reservations'][0]['Instances']  # TODO: fix this
```
silvexis: Do **not** use this method for getting all instances, this is considered the "old fashioned" way to do it and will not return a complete list of instances in most cases where there are lots of EC2 instances. Instead use `instance_data = aws.resource.instances.all()`

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,384 +1,205 @@
...
+        'monitoring_state': instance_data.get('Monitoring', {}).get('State'),
+        'state_code': instance_data.get('State', {}).get('Code'),
+        'state_name': instance_data.get('State', {}).get('Name'),
+        'state_transition_reason': instance_data.get('StateTransitionReason'),
 
-    namespace = os.environ.get('NAMESPACE')
+        # TODO: figure out whether we should treat VPC and AZ as first-class citizens that all resources get
+        # and that are defined in CZ-NRF (i.e. the "basis").  These may be handled better as derived relationships.
+        # Also not sure about what to do for AZ if there's no placement group.
+        'availability_zone': instance_data.get('Placement', {}).get('AvailabilityZone'),
```
silvexis: AFAIK there will always be a Placement group

reactor/features/resources/plugins/ec2/instance.py
```diff
@@ -1,384 +1,205 @@
...
+        'KernelId', 'KeyName', 'LaunchTime', 'VpcId', 'Platform', 'RamdiskId', 'SourceDestCheck',
+        'VirtualizationType', 'EnaSupport', 'Hypervisor',
+    }
 
-    """
+    interesting_properties = {resource.normalize_property_name(k): instance_data.get(k) for k in interesting_keys}
 
-    if not log_group_name:
-        return False
+    other_properties = {
```
silvexis: unsure why you distinguish between interesting and other properties. They all look interesting to me and all get rolled up below.

reactor/features/resources/plugins/ec2/instance.py
silvexis: you should get tenancy from the Placement dict as well

reactor/features/resources/plugins/ec2/instance.py
silvexis: interesting things that are missing from this list: `ImageId`, `PrivateDnsName`, `PrivateIpAddress`, `SecurityGroups`, `SourceDestCheck`

silvexis: even if you pull some of these out as relationship's btw, you should still record them as properties

reactor/features/resources/plugins/ec2/vpc.py
```diff
@@ -0,0 +1,178 @@
...
+            enriched_resource['cloud_vpc_id'] = data['properties']['vpc_id']
+            enriched_resource['resource_state'] = data['properties']['state']
+            enriched_resource['cloud_local_id'] = vpc_id
+            enriched_resource['cloud_global_id'] = None  # VPCs don't have ARNs
+
+            # For future use in CZ-NRF
+            enriched_resource['resource_properties'] = None
+            enriched_resource['resource_relationships'] = None
+            enriched_resource['resource_tags'] = None
+            enriched_resource['resource_present'] = None
```
silvexis: should be True

reactor/features/resources/plugins/plugin_loader.py
```diff
@@ -0,0 +1,43 @@
...
+Module used to manage loading and validation of enrichment plugins.
+"""
+
+import importlib
+import lambda_tools as tools
+
+logger = tools.setup_logging('reactor')
+
+
+_IMPORT_TEMPLATE = 'reactor.features.resources.plugins.{}.{}'
```
silvexis: scope this inside `find_enrichment_plugin`? I find it odd to see it floating outside here when it's not used anywhere else

test/unit/features/plugins/test_plugin_loader.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.mark.enrichment_plugins
+@pytest.mark.unit
+def test_all_plugins_have_cooldown_function(context):
+    for resource_type, service_type in context.resource_type_list:
+        module = importlib.import_module(f"reactor.features.resources.plugins.{service_type}.{resource_type}")
+        assert len(inspect.signature(module.enrich_cooldown).parameters) == 0
+        assert module.enrich_cooldown() > 0
```
silvexis: should be >=

reactor/database/models/dynamodb/cz_reactor_resources.py
```diff
@@ -5,24 +5,22 @@
...
 
 import botocore.exceptions
 from boto3.dynamodb.conditions import Key, Attr
 
 import reactor.constants
-from lambda_tools import setup_logging
+import lambda_tools as tools
```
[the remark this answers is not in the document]
killthrush: Fair enough - it's self-documenting

reactor/features/resources/enrich.py
```diff
@@ -73,7 +76,20 @@ def enrich(event):
...
+    message_source_environment_id = resource_stream_event['environment_id']
+    cloud_account_id = resource_stream_event['cloud_account_id']
+    timestamp = resource_stream_event['timestamp']
+    relationship_resource_ids = set([res_id for relationships in enriched_resources for res_id in relationships])
+    ids_and_timestamps = {id: timestamp for id in relationship_resource_ids}
+    try:
+        resources_dbh.create_new_resources(ids_and_timestamps, message_source_environment_id, cloud_account_id)
+    except Exception as err:
+        logger.error(f'Failed to create {len(relationship_resource_ids)} resources, skipping resource create')
+        logger.exception("{} : {}".format(type(err), err))
```
killthrush: Realized that we never wrote any code to put relationships back into circulation - does this look OK @silvexis ?

reactor/features/resources/plugins/ec2/instance.py
killthrush: This is more for later.  I think what I'd like to do is make the cooldown apply ONLY to these boto calls and not to enrichment generally.  I'm not convinced this is working correctly at the moment.

reactor/features/resources/plugins/ec2/instance.py
killthrush: Hmm... that's getting into a different ticket that's not in the sprint.  I guess I could update these attributes, but I'm not going to mess with the diff logic at all and I'm not going to worry about CZ-NRF until those other tickets get worked on.

reactor/features/resources/plugins/ec2/instance.py
killthrush: Would a terminated instance count as "present"?  If so, I'll make that change

reactor/features/resources/enrich.py
silvexis: it does. It's going to generate a lot of inserts though, I would put a simple throttle on the id's you are writing so that you only write ID's that are new or haven't been seen in last 30 seconds.

reactor/features/resources/plugins/ec2/instance.py
killthrush: damn.  this is going to torpedo everything I've worked on since the data spec is completely different.  I'm going to leave this alone.

reactor/features/resources/plugins/ec2/instance.py
killthrush: Disagree on making relationships properties.  Especially those ones - they are higher-order relationships that should be resolved in Core, not here.  SourceDestCheck I could swear I already had - I'll check that one

reactor/features/resources/plugins/ec2/instance.py
killthrush: There are ones that can be resolved in a very simple fashion and others that need to be plucked from a more complex data structure.  However, I think I could probably use one of those dict util functions here to make this less complicated - I'll try it.

reactor/features/resources/plugins/ec2/instance.py
killthrush: Yeah SourceDestCheck is already there.  RE: the other properties, I will add them temporarily until we have enrichment plugins for the relationships that truly own these properties - the data IS lost if there's no AMI plugin for example.  However I will state again that the strength of our platform will be in the resource graph - all views on our model will ultimately be projections of this graph, and the graph's design is paramount.  If there's value in this sort of summarization (clearly there is - no argument there), then all we need is to build our projections appropriately.  It does not make sense to summarize early simply because that's what AWS has chosen to do.

reactor/features/resources/enrich.py
killthrush: I'll try something in my next PR (ENI plugin).  I'd prefer to have a way to load test this without needing to put this in prod.   hmmm.

PR #80, opened 2018-05-05
Fan out resource writes to S3

silvexis:
### Description
Addresses issues in curate performance during the persisting of resources to S3


reactor/features/resources/__init__.py
```diff
@@ -7,3 +7,4 @@
 from reactor.features.resources.enrich import enrich  # noqa: f401
 from reactor.features.resources.reconcile import reconcile  # noqa: f401
 from reactor.features.resources import plugins  # noqa: f401
+from reactor.features.resources.persist import persist  # noqa: f401
```
killthrush: I take it when my code is merged, this will go away?

reactor/features/resources/curate.py
```diff
@@ -203,17 +200,20 @@ def curate(records, reactor_account_id, namespace):
...
     start = timeit.default_timer()
-    logger.info(f"Writing {len(resources_to_write)} resources out of {event_stats['coresync']} to s3")
+    logger.info(f"Persisting {len(resources_to_write)} resources out of {event_stats['coresync']} to s3")
     for res_id, resource in resources_to_write.items():
-        write_resource_to_s3(res_id, resource['environment_id'], resource['data_image'])
-    processing_stats['coresync'] += (timeit.default_timer() - start)
+        payload = {'res_id': res_id,
+                   'environment_id': resource['environment_id'],
+                   'resource_data': resource['data_image']}
+        lambda_utils.reactor_invoke(namespace, 'resources-persist', payload=json.dumps(payload), async=True)
```
killthrush: After you rebase off my branch, consider using the pyfaaster version of this function instead - one less call site to refactor later

reactor/features/resources/curate.py
```diff
@@ -274,28 +274,6 @@ def process_new_resource(dbh, res_id, timestamp, service_type, resource_type, en
...
-
-    """
-    bucket_name = get_reactor_bucket_name()
-    res_id_hash = hashlib.sha256(res_id.encode('UTF-8')).hexdigest()
-    resource_object_key = f'resources/{environment_id}/{res_id_hash}.gz'
-    resource_object_value = utils.encode_and_compress(json.dumps(resource))
-    logger.debug(f'About to write to the "{bucket_name}" bucket under the key "{resource_object_key}": {resource}')
-    result = Provider(service='s3').client.put_object(Bucket=bucket_name, Key=resource_object_key,
-                                                      Body=resource_object_value, ServerSideEncryption='AES256')
-    return result
```
killthrush: How much faster is the fanout?  I'm guessing that this works because of the async invocation, though I still find this puzzling.  The lambda service itself needs to get all the same payloads we were writing to s3...  Is the encryption and compression the culprit?

reactor/features/resources/persist.py
```diff
@@ -0,0 +1,42 @@
...
+        resource_data.pop(field, None)
+
+    bucket_name = get_reactor_bucket_name()
+    res_id_hash = hashlib.sha256(res_id.encode('UTF-8')).hexdigest()
+    resource_object_key = f'resources/{environment_id}/{res_id_hash}.gz'
+    resource_object_value = utils.encode_and_compress(json.dumps(resource_data))
+    logger.info(f"Persisting '{bucket_name}/{resource_object_key}'")
+    logger.debug(resource_data)
+    Provider(service='s3').client.put_object(Bucket=bucket_name, Key=resource_object_key,
+                                             Body=resource_object_value)  # , ServerSideEncryption='AES256'
```
killthrush: Re-enable encryption before checking this in?

reactor/features/resources/persist.py
```diff
@@ -0,0 +1,42 @@
...
+
+import lambda_tools
+import reactor.utils as utils
+from reactor.aws.provider import Provider
+from reactor.aws.tools import get_reactor_bucket_name
+
+logger = lambda_tools.setup_logging('reactor')
+
+
+def persist(res_id, environment_id, resource_data):
```
killthrush: This function looks superficially to be the same as `write_resource_to_s3` - were there any other changes you remember making?

reactor/features/resources/persist.py
```diff
@@ -0,0 +1,42 @@
...
+        resource_data (dict): Resource data to write
+
+    Returns:
+        tuple: (bucket name, key)
+
+    """
+    # Fields to filter out of the final resource record written
+    filter_fields = ['cz_metadata']
+    for field in filter_fields:
+        resource_data.pop(field, None)
```
killthrush: Ah, this code is different from the original - thanks for fixing this

reactor/features/resources/persist.py
```diff
@@ -0,0 +1,42 @@
...
+
+    bucket_name = get_reactor_bucket_name()
+    res_id_hash = hashlib.sha256(res_id.encode('UTF-8')).hexdigest()
+    resource_object_key = f'resources/{environment_id}/{res_id_hash}.gz'
+    resource_object_value = utils.encode_and_compress(json.dumps(resource_data))
+    logger.info(f"Persisting '{bucket_name}/{resource_object_key}'")
+    logger.debug(resource_data)
```
killthrush: Is this debugger statement more verbose than we want?

requirements.txt
```diff
@@ -1,13 +1,9 @@
...
-botocore==1.9.22
-deprecation==2.0
+boto3==1.7.14
 docopt==0.6.2
-jwt==0.5.3
 PyAthenaJDBC==1.3.3
-simplejson==3.13.2
+simplejson==3.14.0
 retrying==1.3.3
-Pympler==0.5
```
killthrush: I suppose it's OK to remove pympler for now.  It is pretty useful for memory profiling though if we ever need it again

requirements.txt
```diff
@@ -1,13 +1,9 @@
 attrdict==2.0.0
-boto3==1.6.22
-botocore==1.9.22
-deprecation==2.0
```
killthrush: please add deprecation back - this allows us to mark obsolete code and make it more visible at runtime

requirements.txt
```diff
@@ -1,13 +1,9 @@
...
+boto3==1.7.14
 docopt==0.6.2
-jwt==0.5.3
```
killthrush: yup, no more need for jwt

requirements-dev.txt
```diff
@@ -1,22 +1,22 @@
...
+pytest-mock==1.10.0
 tox==3.0.0
-sphinx==1.7.2
+sphinx==1.7.4
 sphinx-autobuild==0.7.1
-sphinx_rtd_theme==0.2.4
+sphinx_rtd_theme==0.3.1
 requests-mock==1.4.0
-pycodestyle==2.3.1
+pycodestyle==2.3.1  # 2.4.0 is unhappy with Flake8 https://gitlab.com/pycqa/flake8/issues/406
```
killthrush: I should add this comment to the other repos

serverless.yml
```diff
@@ -153,10 +153,13 @@ functions:
...
     memorySize: 1024
-  # curate
+  # Resources
+  resources-persist:
+    handler: lambda_handlers.resources__persist
+    memorySize: 512
   resources-curate:
     handler: lambda_handlers.resources__curate
-    memorySize: 2048
+    memorySize: 1024
```
killthrush: In the notes from the AWS tech, it looks like we want the lambda attached to the stream to be as big as it can be for maximum throughput (at least that was how I read it...).  Change this back to 2048?

test/integration/features/resources/test_persist.py
```diff
@@ -0,0 +1,32 @@
...
+
+import simplejson as json
+import boto3
+
+from reactor.aws.tools import get_reactor_bucket_name
+from reactor.features.resources import persist
+from reactor.utils import decompress_and_decode
+
+
+def test_persist_resource(localstack_s3_endpoint):
```
killthrush: *GASP*!  A test! 🦄

reactor/features/resources/curate.py
silvexis: I'm sure that adds to it, but in general s3 writes are slower than a lambda invoke

serverless.yml
silvexis: i did some testing and the performance was nearly identical for any value over 1024

requirements.txt
silvexis: is it being used anywhere? Only reason i removed it is I couldn't find it in use anywhere in the codebase

reactor/features/resources/persist.py
silvexis: no i just changed the name. I renamed the coresync step on the resources flow diagram to "persist" and just brought the code in line

reactor/features/resources/persist.py
silvexis: ooooh! very very good catch

requirements.txt
killthrush: It's a good thing that it's not used at the moment - that means we've cleaned up deprecated constructs.  I'd like to have this available in all our projects though so we can do more than just add TODOs when we want to clean things up.  The overhead is minimal.

PR #82, opened 2018-05-11
REACTOR-378: CZRN Mapping Table

Tankanow:
### Description
This PR adds a CZRN Mapping Table to Serverless and a corresponding model Class.


reactor/database/models/dynamodb/cz_reactor_czrn_map.py
```diff
@@ -0,0 +1,40 @@
...
+
+logger = setup_logging('reactor')
+
+
+class CZReactorCzrnMap(CZReactorModel):
+    _TABLE_BASENAME = 'cz_reactor_czrn_map'
+
+    _TABLE_DEFINITION = {
+        'KeySchema': [{'AttributeName': 'cloud_local_id', 'KeyType': 'HASH'},
+                      {'AttributeName': 'environment_id', 'KeyType': 'RANGE'}],
```
killthrush: Erik raised a good point to me yesterday and that's that IP addresses are not unique even within AWS accounts.  However, if you add ENI to the key then you're good.  They are unique per account.  So I'd propose making a sort key that includes both envid and eni (this is a common dynamo pattern from the looks of it)

reactor/database/models/dynamodb/cz_reactor_czrn_map.py
```diff
@@ -0,0 +1,40 @@
...
+        'AttributeDefinitions': [{'AttributeName': 'cloud_local_id', 'AttributeType': 'S'},
+                                 {'AttributeName': 'environment_id', 'AttributeType': 'S'},
+                                 {'AttributeName': 'czrn', 'AttributeType': 'S'},],
+        'ProvisionedThroughput': {'ReadCapacityUnits': 1, 'WriteCapacityUnits': 1}
```
killthrush: As discussed, we should use a list of maps attribute here - the map would contain `ip` and `czrn` or `ip`, `port`, and `czrn`.  When my action code looks up an IP using address, eni, and env_id, I will fill in all the CZRNs I can

Tankanow: 👍

reactor/database/models/dynamodb/cz_reactor_czrn_map.py
Tankanow: My thinking here is that `cloud_local_id` is a good partition key b/c it is the fastest way to get to a range of data we might care about. For example, if we're looking for `1.1.1.1`, it's better to check that _any_ environment has seen `1.1.1.1` then an `environment_id`. As for uniqueness, I don't really care b/c once the client finds `1.1.1.1` for environment `1`, it can fairly quickly process the `czrn_mappings` list for the `port` (or any other sub-key) it needs.

Tankanow: Yes! I'm advocating for the latter.

reactor/database/models/dynamodb/cz_reactor_czrn_map.py
```diff
@@ -0,0 +1,38 @@
...
+        'AttributeDefinitions': [{'AttributeName': 'cloud_local_id', 'AttributeType': 'S'},
+                                 {'AttributeName': 'environment_id', 'AttributeType': 'S'}],
+        'ProvisionedThroughput': {'ReadCapacityUnits': 1, 'WriteCapacityUnits': 1}
+    }
+
+    def __init__(self, namespace: str = None, aws: Provider = None):
+        super().__init__(self._TABLE_BASENAME, namespace, aws)
+
+    # Note: this caches within a given lambda if you reuse the same object
+    @functools.lru_cache(maxsize=256)
```
killthrush: We have an emerging convention (arguably not an AGREED convention) to use a `_cached` suffix when using the decorator just to make it more obvious to callers.  Thoughts?

reactor/database/models/dynamodb/cz_reactor_czrn_map.py
```diff
@@ -0,0 +1,38 @@
...
+    def get(self, cloud_local_id, environment_id, mappings_filter=None):
+        response = self.table.get_item(Key={'cloud_local_id': cloud_local_id,
+                                            'environment_id': environment_id})
+        mappings = utils.deep_get(response, 'Item', 'czrn_mappings')
+        if mappings_filter and mappings:
+            (k, v) = mappings_filter
```
killthrush: what is `mappings_filter`?  did you intend this to be a predicate function?  This function could use a docstring

test/integration/database/models/dynamodb/test_cz_reactor_czrn_map.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+from reactor.database.models.dynamodb import CZReactorCzrnMap
+
+
+@pytest.fixture(scope='function')
+def mock_logger():
+    patcher = mock.patch('reactor.database.models.dynamodb.cz_reactor_czrn_map.logger')
+    yield patcher.start()
+    patcher.stop()
```
killthrush: you might be able to omit the logger mock in this fixture

test/integration/database/models/dynamodb/test_cz_reactor_czrn_map.py
```diff
@@ -0,0 +1,56 @@
...
+            'environment_id': environment_id,
+            'czrn_mappings': [
+                {'port': 1234, 'czrn': '1'},
+                {'port': 8080, 'czrn': '2'},
+            ]
+        }
+    )
+    czrn_map_dbh = CZReactorCzrnMap(namespace='test')
+    mappings = czrn_map_dbh.get(cloud_local_id, environment_id, mappings_filter=('port', 8080))
+    assert len(mappings) == 1
```
killthrush: Make a test for the caching behavior?  We rely pretty heavily on this in the reactor and it would be good to know if it breaks

reactor/database/models/dynamodb/cz_reactor_czrn_map.py
killthrush: Oh, NVM, now I see how you use it by looking at the tests.  I think this pattern will work, however in real life I will need multiple filters in some situations, i.e. port and ENI

killthrush: I like the function approach

reactor/database/models/dynamodb/cz_reactor_czrn_map.py
killthrush: The `_cached` suffix definitely seems to be the convention in reactor

PR #84, opened 2018-05-14
Feature/reactor 373 - cross-account vpc flowlog subscriptions

killthrush:
### Description

In order to subscribe to VPC flowlogs across accounts, need to set up Log Destinations in concert with kinesis.  Add configuration to configure kinesis, destinations, and supporting roles in the reactor account, and write some code to wire up these destinations in remote accounts.  This code will run when new VPC resources are detected.  Also refactor some details around resource constants and add a default plugin for cz-network resources so they show us as Identified.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -86,7 +116,79 @@ def create_vpc_flow_log(vpc_id, logs_provider, ec2_provider):
         return log_group_name
 
 
-def subscribe_lambda_to_flow_log(log_group_name, logs_client):
+def _subscribe_kinesis_to_flow_log(kinesis_arn, log_destination_role_arn, log_group_name, remote_logs_provider):
```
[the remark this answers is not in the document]
silvexis: I feel your pain

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -86,7 +113,79 @@ def create_vpc_flow_log(vpc_id, logs_provider, ec2_provider):
...
+            'local_account_id': local_account_id,
+            'log_destination_name': destination_args['destinationName'],
+        }
+        logger.debug(f'Setting Destination Policy with args: {policy_args}')
+        destination_policy = _DESTINATION_POLICY_TEMPLATE.format(**policy_args)
+        local_logs_client.put_destination_policy(destinationName=policy_args['log_destination_name'],
+                                                 accessPolicy=destination_policy)
+
+        logger.debug(f'Applying Kinesis subscription filter to {log_group_name}')
+        remote_logs_client.put_subscription_filter(
```
silvexis: Could the situation where the permissions are invalid (on the clients account side) be better handled here? I suspect it's going to be a common occurrence and I'd like to bubble that sort of issue up so we can address it even if it's a manual process today.

Long term we should come up with a strategy to generate CZ events into our own system that can eventually make it all the way to core (and the customer) so our system can tell them what to do/fix when things are misconfigured

reactor/features/resources/plugins/ec2/vpc.py
```diff
@@ -127,23 +120,6 @@ def _get_vpc_relationships(vpc_data, region, account_id):
...
-    flow_log_status = cz_metadata.get('flow_log_status')
-    if not flow_log_status:
-        logger.info(f"Attempting to configure VPC Flow logs for {vpc_id}")
-        flow_log_group_name, subscribe_success = flow_log.configure_flow_log(ec2_provider, vpc_id)
-        flow_log_status = {"log_group_name": flow_log_group_name,
-                           "type": "local",
-                           "subscribed": subscribe_success,
-                           "timestamp": dt.datetime.now(tz=dt.timezone.utc).isoformat()}
-        cz_metadata['flow_log_status'] = flow_log_status
-    return cz_metadata
```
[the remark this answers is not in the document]
silvexis: so the post enrich actions are great, but all of the metadata code should have moved over there too. We do not want the subscribe kinesis to the flow log running every single time we see a VPC, it's going to generate hundreds or thousands of unnecessary API calls and in the case of accounts that lack the permissions for us to subscribe, thousands of errors too. The whole point of the metadata is to capture resource specific state so the system can be smart when it sees a resource for the second, third, or 50 hundredth time.

reactor/features/resources/enrich.py
```diff
@@ -64,7 +65,11 @@ def enrich(resource_stream_event):
         _update_plugin_cooldown_value(enrich_plugin, resource_stream_event)
         logger.info(f"Executing enrichment plugin for {res_id}:{resource_stream_event['timestamp']}")
         resources_to_reconcile = _run_enrichment_plugin(enrich_plugin, resource_stream_event)
-        _record_relationships(resource_stream_event, resources_to_reconcile)
+        try:
+            _record_relationships(resource_stream_event, resources_to_reconcile)
+            _process_actions(resource_stream_event, resources_to_reconcile, configuration)
```
silvexis: I think `_process_actions` should actually live in reconcile and operate on a reconciled resource, because this is going to get called _a lot_. Lets discuss if it doesn't make sense, but if you move `_process_actions` into reconcile, actually make it `_process_action` (drop the s) and then add back metadata processing so that you can avoid running actions that have already run this will all run much more efficiently and safely (In the case you have an action you only want to run 1 time only per vpc, such as the subscribe flow logs action).

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -164,5 +263,6 @@ def get_flow_log_for_vpc_cached(vpc_id, role_arn, external_id, region):
     return flow_logs.get(vpc_id)
 
 
-def get_flow_log_group_name_for_vpc(vpc_id):
-    return f"/cz/vpcflow/{vpc_id}"
+# TODO add reactor id to this
```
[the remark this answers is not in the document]
killthrush: yup!

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -8,70 +8,100 @@
...
         logger.debug(f'Log group {log_group_name} already exists, skipping {error}')
 
-    logs_provider.client.put_retention_policy(logGroupName=log_group_name, retentionInDays=1)
+    remote_logs_provider.client.put_retention_policy(logGroupName=log_group_name, retentionInDays=1)
 
-    existing_flow_log = get_flow_log_for_vpc_cached(vpc_id,
-                                                    role_arn=ec2_provider.role_arn,
-                                                    external_id=ec2_provider.external_id,
-                                                    region=ec2_provider.region)
+    # TODO: include flowlog status when we proactively fetch cloud account information like instances, etc
```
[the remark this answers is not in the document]
killthrush: Oh, I remember - this was my solution to avoiding dynamodb writes after subscribing flowlogs.  This still matters!

reactor/features/resources/actions/vpc_flow_log_action.py
killthrush: Oh, good idea - anytime there's a remote client, that sounds like something we should do

killthrush: I'll add a specialized exception handler for this.  This should, however, probably be part of the full fledged reactor heath check

reactor/features/resources/enrich.py
killthrush: Yeah, I think that makes sense - actions is essentially a step *after* reconcile, once all the records have settled.  I'll see how hard that would be to do.  Regarding the CZ-metadata though, I definitely want to avoid it, starting now if possible.  I understand that you're trying to prevent API calls, but that's not a very good solution because it causes resource writes as a side effect.  One of the things we sorely need to do is make resource operations simple and predictable.  The cached API read we're doing should be more than sufficient for now, and this can be part of the caching strategy we talked about last week.  Thanks what my drunk comment was about - I remembered!

reactor/features/resources/enrich.py
```diff
@@ -102,6 +107,35 @@ def _record_relationships(resource_stream_event, enriched_resources):
...
+        r.get('cloud_local_id') for r in enriched_resources
+        if r.get('service_type') == 'ec2' and r.get('resource_type') == 'vpc' and
+        r.get('disposition') == resource.DISPOSITION_NEW
+    ])
+    logger.debug(new_vpcs)
+    for vpc_id in new_vpcs:
+        vpc_action.configure_flow_log(aws_account_details, vpc_id, configuration)
+
+    # For future use
+    # new_ip_addresses = []
```
killthrush: I'll see about moving this to reconcile instead

reactor/features/resources/plugins/ec2/vpc.py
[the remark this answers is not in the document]
killthrush: We're not in danger of that I don't think - we are performing a single cached lookup when we see a new vpc.  See my previous comments - this is worth a discussion.

lambda_handlers.py
```diff
@@ -199,7 +199,7 @@ def resources__curate(event, context, namespace, configuration, client_details):
 
 @lambda_handler()
 def resources__enrich(event, context, namespace, configuration, client_details):
-    return enrich.enrich(event)
+    return enrich.enrich(event, configuration)
```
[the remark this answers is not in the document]
Tankanow: If it's optional, consider using a kwarg.

reactor/common/czrn.py
```diff
@@ -708,7 +712,7 @@ def extract_all_components(czrn):
     if not czrn_match or len(czrn_match.groups()) != 5:
         raise ValueError(f'CZRN {czrn} did not match standard regex')
     service, region, account_id, resource_type, cloud_local_id = czrn_match.groups()
-    if service not in _service_types_in_arns:
+    if service not in _service_types_in_arns.union(_cloudzero_service_types):
```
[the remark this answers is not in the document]
Tankanow: Might want to do this `union` once at the top of the module rather than in this function.

reactor/constants.py
```diff
@@ -4,7 +4,9 @@
 import enum
 
 DEFAULT_REACTOR_SETTINGS = {'slack_webhook_url': None,
-                            'coresync_enabled': False}
+                            'coresync_enabled': False,
+                            'kinesis_arn': None,
+                            'log_destination_role_arn': None}
```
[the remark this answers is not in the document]
Tankanow: Does the reactor support nested config? If so, consider nesting this stuff under a key like `vpc`. Also, please qualify `kinesis_arn` and `log_destination_role_arn` by their function instead of the service, e.g. `vpc_event_stream_arn`.

reactor/database/models/dynamodb/cz_reactor_resources.py
```diff
@@ -57,9 +56,9 @@ def write(self, records):
                                      'required hash or range key: {}'.format(data))
 
                 if not data.get('disposition'):
-                    data['disposition'] = reactor.constants.RESOURCE_DISPOSITION_PENDING
+                    data['disposition'] = resource.DISPOSITION_NEW
```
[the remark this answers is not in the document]
Tankanow: consider
```
if 'disposition' not in data:
    data['disposition'] = resource.DISPOSITION_NEW
```
or even better
`data['disposition'] = data.get('disposition', resource.DISPOSITION_NEW)`

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -8,70 +8,97 @@
 from lambda_tools import setup_logging
 from reactor.aws.lambda_utils import get_fully_resolved_function_name, get_arn_of_lambda_function_cached
 from reactor.aws.provider import Provider
+from reactor.aws.tools import get_account_id_from_arn, get_region_from_arn
```
Tankanow: Reactor doesn't follow our import conventions. Are we trying to stay consistent in Reactor until we can change them all?

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -8,70 +8,100 @@
...
+                                                               flow_log_group_name, remote_logs_provider)
         else:
             # This attempts to subscribe our VPC Flow Log to a reactor ingest lambda however
             # this method only works for log groups in the same account as the reactor.
-            # For a more general approach, you must use the Kinesis method (optional)
-            # subscribe_lambda_to_flow_log(flow_log_group_name, logs_provider)
-            logger.info("Skipping Lambda Subscription until REACTOR-373")
-            subscribe_success = True
+            # For a more general approach, you must use the Kinesis method (above, optional)
+            subscribe_success = _subscribe_lambda_to_flow_log(flow_log_group_name, remote_logs_provider)
```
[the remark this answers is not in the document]
Tankanow: _This_ code works; however, the VPC ingest code no longer supports Lambda. We should restore the old lambda if we really want to support both Kinesis and Lambda.

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -8,70 +8,97 @@
...
     except Exception as error:
-        logger.warning(f"Could not check or enable flow logs for VPC {vpc_id}, check permissions ({error})")
+        message = f'Could not check or enable flow logs for VPC {vpc_id}, check permissions ({error})'
```
Tankanow: do you want to `logger.exception` here instead of combining the error into the message?

reactor/features/resources/actions/vpc_flow_log_action.py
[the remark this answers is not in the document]
Tankanow: You should add a comment that helps us all remember this point.

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -8,70 +8,97 @@
...
-    logs_provider.client.put_retention_policy(logGroupName=log_group_name, retentionInDays=1)
+    remote_logs_provider.client.put_retention_policy(logGroupName=log_group_name, retentionInDays=1)
 
-    existing_flow_log = get_flow_log_for_vpc_cached(vpc_id,
-                                                    role_arn=ec2_provider.role_arn,
-                                                    external_id=ec2_provider.external_id,
-                                                    region=ec2_provider.region)
+    # TODO: include flowlog status when we proactively fetch cloud account information like instances, etc
+    existing_flow_log = _get_flow_log_for_vpc_cached(vpc_id,
+                                                     role_arn=remote_ec2_provider.role_arn,
```
Tankanow: Consider reusing same `**remote_ec2_provider.aws_connection_details` syntax you used above.

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -86,7 +113,79 @@ def create_vpc_flow_log(vpc_id, logs_provider, ec2_provider):
...
+    local_logs_provider = Provider('logs')
+    local_lambda_provider = Provider('lambda')
+    local_logs_client = local_logs_provider.client
+    remote_logs_client = remote_logs_provider.client
+    local_lambda_client = local_lambda_provider.client
+
+    if not log_group_name:
+        return False
+
+    namespace = os.environ.get('NAMESPACE')
```
Tankanow: Is there a way to avoid this `os` way down here?

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -86,7 +113,79 @@ def create_vpc_flow_log(vpc_id, logs_provider, ec2_provider):
...
+
+    try:
+        destination_args = {
+            'destinationName': log_group_name.strip('/').replace('/', '-'),
```
Tankanow: What's the purpose of this `log_group_name` munging?

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -86,7 +113,79 @@ def create_vpc_flow_log(vpc_id, logs_provider, ec2_provider):
...
+        logger.debug(f'Setting Destination Policy with args: {policy_args}')
+        destination_policy = _DESTINATION_POLICY_TEMPLATE.format(**policy_args)
+        local_logs_client.put_destination_policy(destinationName=policy_args['log_destination_name'],
+                                                 accessPolicy=destination_policy)
+
+        logger.debug(f'Applying Kinesis subscription filter to {log_group_name}')
+        remote_logs_client.put_subscription_filter(
+            destinationArn=destination_arn,
+            filterName='flowlogs',
+            filterPattern='[version, account_id, interface_id, srcaddr != "-", dstaddr != "-", srcport != "-", '
```
Tankanow: TODO: turn this filterPattern into public constant that can be used for testing, etc.

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -152,7 +251,7 @@ def add_lambda_invoke_permissions(function_name, principal):
 
 
 @functools.lru_cache(maxsize=100)
```
Tankanow: `maxsize` [should be power-of-2](https://docs.python.org/3/library/functools.html?highlight=lru_cache#functools.lru_cache).

reactor/features/resources/enrich.py
```diff
@@ -64,7 +65,11 @@ def enrich(resource_stream_event):
         _update_plugin_cooldown_value(enrich_plugin, resource_stream_event)
         logger.info(f"Executing enrichment plugin for {res_id}:{resource_stream_event['timestamp']}")
         resources_to_reconcile = _run_enrichment_plugin(enrich_plugin, resource_stream_event)
-        _record_relationships(resource_stream_event, resources_to_reconcile)
+        try:
+            _record_relationships(resource_stream_event, resources_to_reconcile)
+            _process_actions(resource_stream_event, resources_to_reconcile, configuration)
+        except Exception as e:
+            logger.exception(e)
```
Tankanow: consider adding `logger.warning` here with human readable message.

reactor/features/resources/enrich.py
```diff
@@ -102,6 +107,35 @@ def _record_relationships(resource_stream_event, enriched_resources):
         logger.exception("{} : {}".format(type(err), err))
 
 
+def _process_actions(resource_stream_event, enriched_resources, configuration):
```
Tankanow: Seems like the action itself should implement `_process_actions`. Maybe I'm not understanding. Does any module that wants "actions" have to implement its own `process_actions` loop? You can explain later.

lambda_handlers.py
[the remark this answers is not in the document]
killthrush: Oh, sorry, the configuration dict is required - it's a couple of config keys that are optional.  If they are not part of config, the default behavior is to attempt a direct lambda subscription (which will fail gracefully if going cross-account)

reactor/common/czrn.py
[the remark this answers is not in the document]
killthrush: Makes sense

reactor/constants.py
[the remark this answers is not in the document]
killthrush: Hmm, I'm not sure, but I suspect it's not something to touch until we migrate to pyfaaster.  Having gone through this in pipeline I think that's safest.  I'll add a comment

reactor/database/models/dynamodb/cz_reactor_resources.py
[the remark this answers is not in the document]
killthrush: I'll change this to the second form - good idea

reactor/features/resources/actions/vpc_flow_log_action.py
killthrush: Yeah, Erik pushed back on a couple of items that I thought merited another look, specifically class imports like `Decimal`.  I definitely prefer our chosen style, but there are a few items mainly from common python libs where `from` style is more consistent with community usage.  Trying not to go too crazy with compliance right now to avoid breakage.  The time to do this IMO is a module at a time where we are refactoring and adding test coverage back.

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -5,73 +5,133 @@
...
+
+    Args:
+        aws_account_details (dict):  Information about the connected cloud account (region, ID, role ARN, etc)
+        enriched_resource (dict): An enriched resource record that contains information to process (or skip)
+        configuration (dict): Reactor configuration
 
-def configure_flow_log(ec2_provider, vpc_id, kinesis=False):
+    Returns:
+        dict: A dict of values that will be added to cz_metadata to record the results of the action.
+    """
```
killthrush: This spec is what I plan to follow for other actions.  At this stage, we should only be writing metadata (or nothing) back to the resource record

reactor/features/resources/actions/vpc_flow_log_action.py
```diff
@@ -5,73 +5,133 @@
...
+    resource_type = enriched_resource.get('resource_type')
+    disposition = enriched_resource.get('disposition')
 
     try:
-        logs_provider = Provider(service='logs', **ec2_provider.aws_connection_details)
-        flow_log_group_name = create_vpc_flow_log(vpc_id, logs_provider, ec2_provider)
+        is_vpc_resource = service_type == 'ec2' and resource_type == 'vpc'
+        if not is_vpc_resource or disposition != resource.DISPOSITION_IDENTIFIED:
+            logger.debug(f'Skipping flowlog action for {res_id}, not an eligible resource')
+            return {}
```
killthrush: We're now checking both the eligible resource types and checking the cache inside the flowlog action function - single responsibility.  Either it has work to do or it doesn't, and failures are handled gracefully through the cz_metadata report

reactor/features/resources/actions/vpc_flow_log_action.py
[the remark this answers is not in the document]
killthrush: If you could, send me the SHA for the commit where I can find the old code.  We can discuss as part of our roadmap thingie

reactor/features/resources/actions/vpc_flow_log_action.py
killthrush: I could remove it from this function, however it would need to be in here somewhere, since it's how we build the name of the lambda function.  That name could be built elsewhere but then it would need to be passed around.  No net improvement IMO

reactor/features/resources/actions/vpc_flow_log_action.py
killthrush: I'm building a resource name (destination) that's compliant with our naming conventions.  The source is the name of the log stream which looks more like a path.  I'll add a comment

reactor/features/resources/plugins/ec2/vpc.py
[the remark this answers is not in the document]
killthrush: OK!  Thanks for the debate on this @silvexis - I think we're in a much better place now.  We're not attempting to use  cz_metadata as a distributed cache, but we're not losing this information either.  And we now have a more repeatable way to build new actions, for example the one I need to build in the next ticket.

reactor/features/resources/reconcile.py
```diff
@@ -142,13 +149,45 @@ def reconcile(resource):
             logger.warning(f'Resource {res_id}:{curr_timestamp} changed before we could update it {error}')
         else:
             raise
+    except Exception as e:
+        logger.exception(e)
 
     return processed_count
 
 
+def _process_actions(resource_stream_event, enriched_resource, configuration):
```
Tankanow: This looks good @killthrush. My only concern is that we seem to have a few different ways of triggering "actions" now and I'm starting to get confused how/when/where things can and do happen. In response, I would like to draw out some of this in a Wiki page. Also, I would like to know where the boundaries of our system lie. For example, there are `Provider`s inside the `configure_flow_log` action.

reactor/features/system.py
```diff
@@ -138,6 +138,8 @@ def setup(new_settings, configuration, namespace, account_id):
     setup_results.update(action=action_msg,
                          reactor_name=prepared_settings['reactor_name'],
                          reactor_id=prepared_settings['reactor_id'],
+                         kinesis_arn=prepared_settings['kinesis_arn'],
+                         log_destination_role_arn=prepared_settings['log_destination_role_arn'],
```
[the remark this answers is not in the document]
Tankanow: Did you see my comment re qualifying these names by function rather than service? For example, I think `vpc_event_stream_arn` is clearer (and leaves room for more stream arns)  than the generic `kinesis_arn`. Same goes for `log_destination_role_arn`.

reactor/features/resources/reconcile.py
killthrush: Agreed - the Reactor does not really have domain organization but needs it.  This will take some dedicated effort clearly beyond the scope of this ticket.  The domain modeling we've done is a start, but we've got quite a bit more work to do.

reactor/features/resources/actions/vpc_flow_log_action.py
killthrush: Erik scheduled 30 minutes later in the week to discuss this, and we could probably take a look at a few other items in our list if we have time

PR #88, opened 2018-05-24
REACTOR-389 - Remove SimpleQuery and extra replication steps from the Reactor

tiyberius:
### Description

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


README.md
```diff
@@ -6,8 +6,6 @@
 
 ---
 
-- Have you ever wanted to quickly query everything going on inside your
-AWS account from one command line?
```
killthrush: Hmm.... I think we may want to do more than just remove this line.  I think we need to rework the described philosophy of the reactor in here.  Its value nowadays is not in queries, but in data normalization via CZ-NEF and CZ-NRF.  Technically you could use reactor to build your own tools - we just so happen to have a nice product that we'd be happy to sell you =)

serverless.yml
```diff
@@ -275,13 +275,6 @@ functions:
               paths:
                 id: true
   # Feature APIs
-  query:
-    handler: lambda_handlers.simplequery__query
-    events:
-      - http:
-          path: features/query
-          method: get
-          private: true
```
killthrush: I think there's more in here that needs to be deleted - there's a bucket, and I think there are permissions and things related to athena (glue?).  Also there is a 'finalizer' type lambda and an SNS message that manages athena resources that we should clean up

swagger.yml
```diff
@@ -345,23 +345,6 @@ paths:
...
-        - in: query
-          name: q
-          type: string
-          required: true
-          description: "The query to be issued to the Reactor. \n\n Example: SELECT * FROM events LIMIT 10;"
-      responses:
-        200:
-          description: OK
-      tags:
-        - "features"
```
killthrush: Are we building swagger docs with semaphore?  How are these updated on swaggerhub?

reactor/features/system.py
```diff
@@ -177,9 +176,7 @@ def setup_finalizer(account_id, namespace, configuration):
     """
     slack = SlackWeb(configuration.get('slack_webhook_url'))
 
-    simple_query_init_result = reactor.features.simplequery.initialize(namespace, configuration.region)
-
-    results = {"SimpleQuery": simple_query_init_result}
+    results = {}
 
     final_results = {}
     for k, v in results.items():
```
imcgaunn: if we set results to `{}` on 179 there isn't a reason to loop through its items on line 182.

serverless.yml
imcgaunn: I would also recommend checking up on athena / glue related resources.

swagger.yml
silvexis: We aren’t doing anything automated with swagger

serverless.yml
silvexis: Also no reason to write events/resources to reactor bucket

README.md
tiyberius: Created https://cloudzero.atlassian.net/browse/REACTOR-396

PR #102, opened 2018-06-08
REACTOR-382 - HOTFIX for everything is an outside resource

tiyberius:
### Description

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


reactor/features/resources/enrich.py
```diff
@@ -128,6 +130,8 @@ def enrich_resource_observations(resource_observations, namespace, reactor_accou
         resource_record = deserialize_dynamodb_stream_event_record(resource_observation)
         resource_id = get_resource_id_from_resource_record(resource_record)
         cz_account_id = get_cz_account_id_from_resource_record(resource_record)
+        environment = _get_environment_for_cz_account_id_cached(cz_account_id)
+        cloud_account_id = get_cloud_account_id_from_environment(environment)
```
killthrush: Instead of importing and using a private function, you should just pass in `account_context` and use its `connected_account_id` attribute (yes, the name is not quite right but the value is what you need).  In addition, you can get `configuration` and `namespace` from the context, so you don't need the extra args.  See `account_context.py` to see all the attributes

killthrush: Overall, the fix on 143 looks good!

tiyberius: Ah, freaking duh. Will fix

PR #116, opened 2018-06-14
Hotfix/reactor 424 - Make account bucket checks more resilient

killthrush:
### Description

Make account bucket checks more resilient to changes in storage class.
Previously, in situations where our cloudtrail bucket has certain
lifecycle rules, some logs might end up in Glacier but still appear as
if they are readable as regular keys.  This would cause the account
check to report failure when there is nothing wrong.

Fix this specific issue by attempting to read only newer logfiles using
the current date.  These can more or less be guaranteed to be readable.

### Acceptance checklist

- [ ] Feedback addressed
- [ ] All tests passed


lambda_handlers.py
```diff
@@ -61,13 +61,6 @@ def system__coresync(event, context, namespace, configuration, client_details):
     return generate_response(200, json.dumps(result))
 
 
-@lambda_handler()
-def system__verify_bucket(event, context, **kwargs):
-    event_body = json.loads(event['body'])
-    account_id = get_account_id(context)
-    return accounts.verify.verify_bucket(event_body, account_id)
```
killthrush: This handler also had the same basic bug.  But upon reviewing the code more closely Erik and I realized that not only is it doing a write check for cloudtrail (huh?)  it's also not being used anymore.  Deleting it

reactor/aws/tools.py
```diff
@@ -120,21 +120,64 @@ def get_region_from_arn(arn):
...
+        for prefix in prefixes:
+            next_result = paginator.paginate(Bucket=bucket_name, Delimiter='/', Prefix=prefix)
+            res = iterate_folders(bucket_name, next_result, depth - 1)
+            next_prefixes |= res
+
+        prefixes |= next_prefixes
+        return prefixes
+
+    folders = iterate_folders(bucket_name, result, depth)
+    return folders
```
killthrush: I didn't end up using this nifty function that Erik wrote in pipeline (was not a great fit for the problem), but it's here and it's got tests! We can probably move to pyfaaster at some point.

reactor/aws/tools.py
```diff
@@ -345,47 +387,6 @@ def naive_verify_role_permissions(aws, role_arn, values_to_check):
...
-        if not cloudtrail_bucket_access:
-            reason = f"Unable to access remote cloudtrail bucket using role {account_details['remote_cloudtrail_role']}"
-            logger.info(f'verify_cloudtrail_bucket_access | results: {cloudtrail_bucket_access} : {reason}')
-    else:
-        cloudtrail_bucket_access, reason = verify_bucket_read(aws, bucket=account_details['cloudtrail_bucket'],
-                                                              prefix=account_details.get(
-                                                                  'cloudtrail_bucket_prefix') or None)
-
-    logger.debug(f'verify_cloudtrail_bucket_access | results: {cloudtrail_bucket_access} : {reason}')
-    return cloudtrail_bucket_access, reason
```
killthrush: These are very specialized, so moved them to Verify.py.  there's no re-use value here

reactor/features/accounts/account_context.py
```diff
@@ -62,7 +68,7 @@ def build_cloud_account_context(namespace, configuration, reactor_account_id,
     context.aws_provider = Provider(role_arn=account_details['role_arn'],
                                     external_id=account_details['external_id'],
                                     region=account_details['region'])
-    logger.debug(f'Built context for account {cloud_account_id or cz_account_id}')
+    logger.info(f'Built context for account {cloud_account_id or cz_account_id}')
```
killthrush: added better logging.  Since this information feeds into a lot of reactor functions now, we'd like to immediately see problems.

reactor/features/accounts/facades/cloudtrail_bucket.py
```diff
@@ -0,0 +1,136 @@
...
+    This set does not need to be exhaustive or even accurate - this is only used as a performance optimization.
+
+    Returns:
+        set - a set of region names that can appear in cloudtrail
+    """
+    return {
+        'us-east-1', 'us-east-2', 'eu-west-1', 'eu-west-2', 'eu-west-3', 'us-west-1', 'us-west-2',
+        'ap-northeast-1', 'ap-northeast-2', 'ap-northeast-3', 'ap-south', 'ap-southeast-1', 'ap-southeast-2',
+        'ca-central-1', 'eu-central-1', 'sa-east-1'
+    }
```
killthrush: New module written as a facade to the cloudtrail bucket.  This is specifically for stuff related to actual boto calls like permission checks, and later for log downloads (this code exists elsewhere)

reactor/features/accounts/facades/cloudtrail_bucket.py
```diff
@@ -0,0 +1,136 @@
...
+        cloud_account_id (str): The (optional) 12-digit ID of the connected AWS account for which to check logs
+
+    Returns:
+        tuple(str, str) - A status code for the operation w/ a descriptive message
+    """
+    if not bucket_name:
+        return ResourceAccess.NOT_OK, 'Bucket name was not supplied'
+
+    try:
+        prefixes = _get_active_log_prefixes(aws, bucket_name, cloud_account_id)
```
killthrush: the basic idea is this - if we can find keys that are really new, we're more or less guaranteed to escape lifecycle issues.  It's a bit tricky though, since we need not only the cloudtrail log prefix (it can be anywhere in the bucket!) - we also need to recurse far enough into the directory using account ID, region and date.  But it's pretty easy once that's out of the way.

reactor/features/accounts/facades/cloudtrail_bucket.py
```diff
@@ -0,0 +1,136 @@
...
+    aws.service = 'cloudtrail'
+    trails = _describe_trails(aws)
+    aws.service = previous_service
+
+    cloudtrail_prefix = None
+    matching_trails = [trail['S3KeyPrefix'] for trail in trails if trail['S3BucketName'] == bucket_name]
+    if matching_trails:
+        cloudtrail_prefix = matching_trails[0]
+
+    return cloudtrail_prefix
```
killthrush: Using this function, we can eliminate the need to collect the cloudtrail prefix from the user - ever!

reactor/features/accounts/verify.py
```diff
@@ -58,70 +62,6 @@ def verify_account_details(configuration, account_details, client_details, react
...
-            aws, bucket=request['cloudtrail_bucket'])
-
-        logger.info(f"verify_bucket | {result} : {reason}")
-        return lambda_tools.generate_response(200, json.dumps(result, cls=jsontools.EnumEncoder))
-    except Exception as err:
-        logger.exception(err)
-        logger.error(request)
-        return lambda_tools.generate_response(500, json.dumps({'error': str(err)}))
-
-
```
killthrush: remove obsolete code

reactor/features/accounts/verify.py
```diff
@@ -180,11 +120,8 @@ def _do_verification(details):
 
             # These _must_ work if anything is to work
             verify_responses['target_account'] = list(aws_tools.verify_account_access(aws))
-            verify_responses['cloudtrail_account'] = list(aws_tools.verify_cloudtrail_access(details))
-
-            # HACK ALERT! work around glacier problem until REACTOR-424 is resolved
-            verify_responses['cloudtrail_bucket'] = [ResourceAccess.OK, 'Because I said so']
-            # verify_responses['cloudtrail_bucket'] = list(aws_tools.verify_cloudtrail_bucket_access(aws, details))
```
killthrush: Erik's hack is no longer needed

reactor/features/accounts/verify.py
```diff
@@ -224,7 +161,8 @@ def _do_verification(details):
     except Exception as error:
         logger.exception(error)
         verify_result.update(
-            {'summary': verify_responses, 'success': False, 'reason': str(error), 'attempts': attempts})
+            {'summary': verify_responses, 'success': False, 'full_success': False,
```
killthrush: fix bug - if an exception was handled, the `full_success` key was missing, causing more errors

reactor/features/accounts/verify.py
```diff
@@ -234,6 +172,42 @@ def _do_verification(details):
...
+        if not cloudtrail_bucket_access:
+            reason = f"Unable to access remote cloudtrail bucket using role {account_details['remote_cloudtrail_role']}"
+            logger.info(f'verify_cloudtrail_bucket_access | results: {cloudtrail_bucket_access} : {reason}')
+    else:
+        cloudtrail_bucket_access, reason = cloudtrail_bucket.verify_read(aws,
+                                                                         account_details['cloudtrail_bucket'],
+                                                                         account_details['connected_account_id'])
+
+    logger.debug(f'verify_cloudtrail_bucket_access | results: {cloudtrail_bucket_access} : {reason}')
+    return cloudtrail_bucket_access, reason
```
killthrush: these were transplanted from `aws_tools` and refactored to call the new cloudtrail bucket module

serverless.yml
```diff
@@ -94,10 +94,13 @@ custom:
...
       prod: "rate(15 minutes)"
       default: "rate(5 minutes)"
+    accountConfigCheck:
+      prod: "rate(1 hour)"
+      default: "rate(1 minute)"
   streams:
     observations:
       batch_size:
-        prod: 25
+        prod: 250
```
killthrush: after looking at prod, realized that the old setting was better.  This is totally unrelated to the bugfix.

serverless.yml
```diff
@@ -345,7 +348,10 @@ functions:
     handler: lambda_handlers.accounts__check_accounts
     events:
       - schedule:
-          rate: rate(1 hour)
+          rate: ${
+                self:custom.cron.accountConfigCheck.${self:provider.stage},
+                self:custom.cron.accountConfigCheck.default
+              }
```
killthrush: this is now configurable.  dev feedback is rapid

reactor/features/accounts/facades/cloudtrail_bucket.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Returns:
+        list(str) - a list matching prefix or None if one could not be found
+    """
+    now = datetime.now(timezone.utc)
+
+    cloudtrail_prefix = _get_cloudtrail_prefix(aws, bucket_name)
+    cloudtrail_regions = _get_cloudtrail_region_list()
+    candidate_current_log_prefixes = [
+        f'{cloudtrail_prefix}/AWSLogs/{cloud_account_id}/CloudTrail/{region}/{now.year}/{now.month:02}/{now.day:02}/'
```
silvexis: it's rare to have a prefix, but this seems to expect it always, is that right?

killthrush: Good catch - I needed to handle the cases where there's no cloudtrail prefix

killthrush: should be fixed now

